Issue #8, opened 2016-09-08
Regression testing against typical web applications


gregw:
regular builds should be tested against deployments of several common webapps such as [spring pet clinic](https://github.com/spring-projects/spring-petclinic), to both demonstrate how to deploy those webapplication into gcloud flex and to test for regressions

meltsufin:
@gregw This is a high priority for us. Can someone on your side help us with this? Running these tests locally would be sufficient for now. We can later integrate them into the CI server, when that's ready.

joakime:
What do you mean by "running these tests locally"?

Due to the nature of how gcloud operates, we cannot run locally any of these images, they have to be deployed to gcloud to test against.

The issue Greg filed was about regression testing, in an automated fashion, several known webapps.

joakime:
Webapps present on `getting-started-java` (once you fix the currently broken build on that repo)

- ./bookshelf/2-structured-data/target/bookshelf-2-1.0-SNAPSHOT.war
- ./bookshelf/3-binary-data/target/bookshelf-3-1.0-SNAPSHOT.war
- ./bookshelf/4-auth/target/bookshelf-4-1.0-SNAPSHOT.war
- ./bookshelf/5-logging/target/bookshelf-5-1.0-SNAPSHOT.war
- ./bookshelf/6-gce/target/bookshelf-6-1.0-SNAPSHOT.war
- ./bookshelf/optional-container-engine/target/bookshelf-gke-1.0-SNAPSHOT.war
- ./helloworld-compat/target/helloworld-compat-1.0-SNAPSHOT.war **(appengine-web.xml present)**
- ./helloworld-jsp/target/helloworld-jsp-1.0-SNAPSHOT.war
- ./helloworld-servlet/target/helloworld-servlet-1.0-SNAPSHOT.war

Do you want all of these presented as dockerized + gcloud deployables for flex only? (leaving out the `helloworld-compat` webapp)

Which repo would you like to see these in?

meltsufin:
What I meant by "running locally" is that instead of having a CI server run the test script, we will invoke it from our personal machines. The tests themselves would involve running the images on GCP. We can then utilize the same tests for regression testing by adding them to a CI server.

See https://github.com/aslo/java-runtimes-test for the kind of thing I have in mind.

meltsufin:
Let's start out with hello-world-servlet (exclude compat). We can add the more complex apps later when the testing pattern is established.
The tests should go into this repo.

[...]

PR #17, opened 2016-09-12
Issue #8 - (phase 1) script based war deployment and testing

joakime:
This is phase 1 for the automatic deployment of webapps and integration testing against the jetty9 base image.

More webapps and cleaner integeration testing will be subsequent phases.


joakime:
Consider this script and demo webapp just a P.O.C

tests/run.sh
```diff
@@ -0,0 +1,41 @@
...
+docker tag "jetty9:latest" "gcr.io/${gcloud_projectid}/jetty9:testing"
+gcloud docker push "gcr.io/${gcloud_projectid}/jetty9:testing"
+
+# Deploy and Execute the individual test-war-* tests
+
+for module in test-war-*
+do
+  echo "module: $module"
+  appengine_deploy "${module}"
+done
```
aslo: It might be a good idea to remove the deployed version after the tests have run.
```
gcloud app versions delete "${test_runid}"
```

tests/test-war-hello/src/test/java/com/google/cloud/runtimes/jetty/tests/DeploymentITCase.java
```diff
@@ -0,0 +1,61 @@
...
+        {
+            IO.copy(reader, writer);
+            return writer.toString();
+        }
+    }
+    
+    @Test
+    public void testGet() throws IOException
+    {
+        HttpURLConnection http = (HttpURLConnection) serverURI.resolve("/hello/").toURL().openConnection();
```
aslo: Do you envision that all test webapps will implement all the same routes?

tests/run.sh
joakime: Noted.

Do you know how to create an machine/instance (or whatever the correct name is) with a specific environment configurations? (like say 2GB or 5GB for the webapp?)

tests/test-war-hello/src/test/java/com/google/cloud/runtimes/jetty/tests/DeploymentITCase.java
joakime: This is just a P.o.C.

I plan on using maven for it all (no shell scripts), and the integration-testing phase across each test-webapp, and will share a common testing core.

tests/run.sh
aslo: It's possible to tweak the VMs via some settings in app.yaml. [See docs](https://cloud.google.com/appengine/docs/flexible/java/configuring-your-app-with-app-yaml). Is that what you mean?

joakime: I'll try the `app.yaml` approach ...

``` yaml
resources:
  cpu: .5
  memory_gb: 1.3
  disk_size_gb: 10
```

This would be good for testing the `GAE_MEMORY_MB` issue as well.

aslo: True, though if you just wanted to test that the jetty runtime obeys the `GAE_MEMORY_MB` env variable, you could just override that variable with an arbitrary value in a custom dockerfile, and then verify that the jvm was started with max heap set to 80% of that value. 

I don't think we need to worry about testing that the `GAE_MEMORY_MB` value is set properly for various VM configurations; I would consider that to be on the app engine platform's side of the contract. @meltsufin Do you have any thoughts on this?

meltsufin: I agree that we don't need to verify the value of GAE_MEMORY_MB because that may change. We just want to make sure that it is set, and that we are assigning 80% of it to JVM heap.

meltsufin:
I think this is on the right track continuing what @aslo started. I like the actual tests being in Java and Maven rather than shell scripts, as long as it can be done with a reasonable amount of effort. Ultimately, we may be able to get by without using gcloud CLI at all and use APIs directly for interacting with cloud services.

tests/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            <groupId>org.apache.maven.surefire</groupId>
+            <artifactId>surefire-junit47</artifactId>
+            <version>2.19.1</version>
+          </dependency>
+        </dependencies>
+      </plugin>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>appengine-maven-plugin</artifactId>
+        <version>0.1.1-beta</version>
```
meltsufin: The latest version is 0.1.2

joakime: That will be a separate PR for issue #22

tests/run.sh
joakime: Point is, we need to test the environment behavior as well.   If we deploy in flex or not, the behavior is different, and is dependent on the memory_gb setting, so having at least the 4 different memory configurations is a good thing to test and ensure that we dont have a regression (which is the root goal of this issue and initial PR)

meltsufin:
Is Alex's comment on `gcloud app versions delete "${test_runid}"` the only thing we're blocking on before committing this?

joakime:
The versions delete will come in with subsequent commit / pull request

meltsufin:
If the app doesn't successfully come up, will the Maven build fail?

meltsufin:
Closing in favor of #45.

Issue #4, opened 2016-09-07
Logging using the Google Cloud Logging API


meltsufin:
The [Google Cloud Logging API](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-logging) or the stdout and stderr should be used to send logs to Stackdriver Logging.

meltsufin:
Would it make sense to start out by making sure that the Jetty server log is directed to standard error? Standard out/error should automatically be picked up by Stackdriver Logging.

joakime:
we would still need java.util.logging, as without it, nothing from the google layers would ever show up in the stackdriver logs without it.

the default jetty StdErrLog is multi-line, which the Stackdriver has had problems with in the past, hence all of the effort to make logging more stable with the json format, etc..

to fix StdErrLog would mean a fork of StdErrLog to something like GoogleStackdriverSystemErrLog just to conform to the limits of the Stackdriver standard out/error support

that effort would be better spent in just implementing the java.util.logging layer properly to Google Cloud Logging.

meltsufin:
> we would still need java.util.logging, as without it, nothing from the google layers would ever show up in the stackdriver logs without it.

What do you have in mind when you say "google layers"?

> the default jetty StdErrLog is multi-line, which the Stackdriver has had problems with in the past, hence all of the effort to make logging more stable with the json format, etc..

There's work going on to automatically detect multi-line java stack traces in the logs. Is there anything else that's potentially multi-line in the Jetty server logs?

> to fix StdErrLog would mean a fork of StdErrLog to something like GoogleStackdriverSystemErrLog just to conform to the limits of the Stackdriver standard out/error support

What limits are you referring to?

> that effort would be better spent in just implementing the java.util.logging layer properly to Google Cloud Logging.

There are java.util.logging handlers in the [Logging API](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-logging/src/main/java/com/google/cloud/logging). Beyond those handlers, what would our logging layer need to do?

joakime:
Multiline output it **very common**, and not just for stack traces.

JSP produces multi-line, JSR365 (javax.websocket) does, most datasources, Jersey, Spring, and even the javax.servlet API does too (example: exceptions during startup / deployment are collected into a multi exception which is a single logging event that reports the main deployment concern, and then lists each deployment failure and possible stacktrace for that failure as well)

[...]

PR #19, opened 2016-09-14
Issue #4 Revert to default jetty logging

gregw:
Reverted to default jetty logging.  Tested with [async-rest](https://github.com/jetty-project/java-docs-samples/tree/gcloud-jetty-runtime-4/managed_vms/async-rest) example, which is currently running at https://jetty9-work.appspot.com/ and the logs produced can be seen on the [console](https://console.cloud.google.com/logs/viewer?project=jetty9-work&authuser=1&key1=default&minLogLevel=0&expandAll=false&resource=appengine.googleapis.com%2Fmodule_id%2Fdefault%2Fversion_id%2F20160914t162607&logName=projects%2Fjetty9-work%2Flogs%2Fappengine.googleapis.com%252Fstderr&lastVisibleTimestampNanos=1473835164000000000).


gregw:
see also https://github.com/GoogleCloudPlatform/java-docs-samples/pull/338

meltsufin:
What will happen to Jetty logging if the `resources/jetty-logging.properties` is not there at all?

joakime:
It would be discovered.

 1. Slf4jLog - used if slf4j-api is present on server side
 2. StdErrLog - used if slf4j-api is not present
 3. JavaUtilLog - never used

It would also make configuring it strange for people that want to change the behavior.
(_"where do I enable log4j? where can I set the logging levels? how can i see the stacktraces?"_ etc)

joakime:
Would recommend leaving `resources/jetty-logging.properties` in place.

It's going to be required in the long run anyway.  

Once any of the following becomes true:

 1. We use the Google Cloud Logging API
 2. We want to use java.util.logging
 3. We want to support `logging.properties` on the application deployment

This PR should be considered "a short term hack".

meltsufin:
My question was more about the lines that are staying in jetty-logging.properites. Are they changing any defaults?
```
org.eclipse.jetty.LEVEL=INFO
org.eclipse.jetty.SOURCE=false
org.eclipse.jetty.STACKS=true
org.eclipse.jetty.LONG=false
```

gregw:
@meltsufin those are the defaults, so yes we can just delete the file, as it is essentially documentation. Any change to the file needs to replace the file in a subsequent docker image, and I have an example of doing that in [the sample app](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/338)

I'll remove

meltsufin:
LGTM

meltsufin:
@gregw I noticed that `java-util-logging.properties` is still there. Was it supposed to be removed as part of this PR?

gregw:
@meltsufin yes that file is no longer used.  Originally it was used if the `jetty-logging.properties` configured JUL logging, but as we have removed `jetty-logging.properties`  we should also remove `java-util-logging.properties`

Issue #20, opened 2016-09-14
remove the docker label prefix 


gregw:
The docker label prefix mechanism is no longer needed.  This was used to add a branch prefix to docker labels, but there is no longer an async branch.

gregw:
@meltsufin can you check the 3 PRs for this issue.

PR #21, opened 2016-09-14
Remove async prefix

gregw:
https://github.com/GoogleCloudPlatform/jetty-runtime/issues/20
parameterized docker image tags and labels for ease of CI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,8 @@
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <jetty.version>9.3.8.v20160314</jetty.version>
     <appengine.api.version>1.9.40</appengine.api.version>
-    <docker.label.prefix>async-</docker.label.prefix>
+    <docker.tag>${jetty.version}</docker.tag>
+    <docker.openjdk.image>openjdk8:8-jre</docker.openjdk.image>
   </properties>
 
   <developers>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -35,7 +35,8 @@
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <jetty.version>9.3.8.v20160314</jetty.version>
     <appengine.api.version>1.9.40</appengine.api.version>
-    <docker.label.prefix>async-</docker.label.prefix>
+    <docker.tag>${jetty.version}</docker.tag>
```
meltsufin: I think the pattern we agreed on is "9.3-${build_id}", where the build_id is a timestamp of the build.

gregw: Ah yes - forgot that.  
So should we change the project version from 0.1.0-SNAPSHOT   to buildID-SNAPSHOT  and just use that?

meltsufin: I'm not really concerned about the Maven artifact version right now. I think we still need to figure out if we want to be publishing to Maven Central at all.

gregw: Then I'd recommend making the artifact version the build ID.  we then get auto increment whenever a release is done - no matter where that is released to.

meltsufin: We're trying to be consistent with other language runtimes, and they seem to be adopting the timestamp `YYYY-MM-DD_HH_MM` as the most specific docker version tag.

PR #40, opened 2016-09-16
Do not create log directory

gregw:
Do not create the log directory, instead defer to openjdk-runtime.
See https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/12


gregw:
@meltsufin please review

meltsufin:
LGTM. Note that the openjdk-runtime may not do this either because we may not be scanning that directory by fluentd anymore.

meltsufin:
Ignore my comment about openjdk-runtime not needing the logging directory. I saw that the Cloud Debugger still needs it.
This one is good to merge.

Issue #62, opened 2016-10-13
Remove Jetty Request Logging


meltsufin:
Followup on #60.

PR #63, opened 2016-10-13
Remove request log configuration #62

gregw:
Remove request log configuration #62

### DIFF
--- a/jetty9-base/src/main/jetty-base/etc/gae.xml
+++ b/jetty9-base/src/main/jetty-base/etc/gae.xml
@@ -27,29 +27,6 @@
     </Set>
   </Ref>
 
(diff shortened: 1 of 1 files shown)


meltsufin:
I'm trying to test this change, but keep running into an error reading root.war.
I have this add instruction:
```
ADD testapp.war $JETTY_BASE/webapps/root.war
```
And I get this error because root.war is owned by root, not jetty user.
```
java.io.FileNotFoundException: /var/lib/jetty/webapps/root.war (Permission denied)
```
I'm not sure why this error doesn't occur when using currently deployed `gcr.io/google_appengine/jetty` image.

I'll come back to this next week to investigate further.

PR #65, opened 2016-10-17
Use the gcr.io/google_appengine/openjdk as base instead of openjdk:8

### LINKED ISSUE #64: Use the gcr.io/google_appengine/openjdk as base instead of openjdk:8

meltsufin:
Fixes #64 

@aslo


meltsufin:
Good catch about the readme. I'll update it as part of this PR.

PR #70, opened 2016-11-15
Always pull latest openjdk base image when building

meltsufin:


gregw:
This PR is necessary as we currently have dependency on a non specific image: `gcr.io/google_appengine/openjdk`, which can change over time, so we cannot allow it to be cached locally.

However, generally speaking I think having non-reproducible builds is not a good thing, specially if you are rebuilding to try to diagnose a problem and a different base image may suddenly appear.

So I would much rather depend on a specific label and save the time/bandwidth by allowing the local image cache to be used.

meltsufin:
@gregw I agree with your comments regarding build mutability, but for now since we're going with the 'latest' tag, we need this. We can revisit both aspects in another PR.

aslo:
:+1:

PR #75, opened 2016-11-23
local and remote testing framework

jmcc0nn3ll:
All tests are located under a /tests directory in the jetty-runtime project.  Each group of tests should exist within the scope of a single maven artifact which contains a deployable docker container (local, remote, or both) and the associated test cases which follow the conventions detailed below.

NOTE: The property ‘jetty.test.image’ will be passed in which is analogous to the FROM Docker directive.  This serves as an effective disconnect between the jetty-runtime project and the specific docker image/tag that the respective tests will be run against.

> cd tests;
> mvn install -Premote -Djetty.test.image=gcr.io/<project>/jetty9:9.3

This will activate a profile and enable remote testing. For each test artifact the appengine-maven-plugin will be used to deploy an instance of the application and the appropriate test cases will be run.  The webapps that will be built will be done through the cloud builder mechanism so the image to be tested (as linked in the jetty.test.image property) will need to be deployed to the appropriate gcr.io location.

> cd tests;
> mvn install -Plocal -Djetty.test.image=jetty:9.x

This will activate a profile and enable local testing.  The tests activated under this profile will make use of the locally installed image and tag.  The docker-maven-plugin is used to build the test docker container and the io.fabric8 plugin is used to manage the integration test lifecycle..  Local tests have a much smaller scope as they are not able to make use of many of the features of the jetty-runtime project.


pom.xml
```diff
@@ -137,7 +138,7 @@
           <version>2.10</version>
         </plugin>
 
-	<plugin>
+	    <plugin>
```
meltsufin: Indentations off?

jmcc0nn3ll:
hm, 503 from the test itself?  the image is being created ok using that gcr.io/google-engine/jetty line right? typically I have to use jetty:9.3 for the local image, it won't download the one it needs from a remote afaik

jmcc0nn3ll:
figure it out, need to increase the time a bit on the test to give the container time to properly startup

fixing an issue with the lifecycle not stopping the docker container in post-integration-test phase before verify and then I'll push a new branch for issue-72 and drop this PR (since it got messed up with the update and picked up TESTING.md)

PR #77, opened 2016-11-25
Set securePort for HTTPS redirects to 443

### LINKED ISSUE #76: Set jetty.httpConfig.securePort=443 for redirects with CONFIDENTIAL security constraints
If a user makes a non https request to a context with a CONFIDENTIAL security constraint, jetty will perform a redirect to the configured secure port.  By default, the value of the port is 8443.  This should be set to 443 instead.

janbartel:
Fixes #76.

This PR sets the securePort to 443 so that redirects can work when a non https request is received on a resource with a CONFIDENTIAL security constraint.

NB the technique of replacing jetty properties in .ini files using the maven-replacer-plugin is something that we can make more use of to customize the basic jetty distro for flex and eventually compat.

### DIFF
--- a/jetty9-base/pom.xml
+++ b/jetty9-base/pom.xml
@@ -139,7 +139,29 @@
           </arguments>
         </configuration>
       </plugin>
-
+      <plugin>
+        <groupId>com.google.code.maven-replacer-plugin</groupId>
+        <artifactId>maven-replacer-plugin</artifactId>
+        <version>1.3.5</version>
+        <executions>
+            <execution>
+                <id>replaceTokens</id>
+                <phase>pre-integration-test</phase>
+                <goals>
+                    <goal>replace</goal>
+                </goals>
+            </execution>
+        </executions>
+        <configuration>
+            <file>target/jetty-base/start.d/server.ini</file>
+            <replacements>
+                <replacement>
+                    <token># jetty.httpConfig.securePort=8443</token>
+                    <value>jetty.httpConfig.securePort=443</value>
+                </replacement>
+            </replacements>
+        </configuration>
+      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
@@ -159,6 +181,8 @@
           <appendAssemblyId>false</appendAssemblyId>
         </configuration>
       </plugin>
+
+
     </plugins>
   </build>
 </project>


meltsufin:
Regarding  the use of the maven-replacer-plugin, it looks like a very indirect way of configuring our Jetty, which can be confusing to people reading the code. Can we just define these configuration properties directly in the src files?

meltsufin:
Regarding the actual change, this does fix the http -> https redirect in my testing.

gregw:
@meltsufin this was discussed a bit by @janbartel and myself.

It would perhaps be clearer to just have the `start.d/server.ini` file in `src` hierarchy and then copy it in.   However, that would not well future proof the build as if other unrelated things change in the server.ini file, these we be lost as it will not be generated.

The other option we discussed was to find a plugin or technique that at least had the replacements in a file held within src rather than in the `pom.xml`, however the other `start.d` contents are created by the `--add-to-start` command in the `pom.xml`, so we felt that it at least kept all the jetty configuration together.

Actually, I have an idea of how we might be able to change the jetty release a little bit to support a file that would contain both the `--add-to-start` command and any property replacements.   

What I'm thinking is that in the pom we just run a command like:
```
java -jar start.jar --configure=src/main/jetty/jetty.configure
```
the referenced file would then be full of lines like:
```
--add-to-start=server jetty.httpConfig.securePort=443
--add-to-start=http-forwarded
--add-to-start=http jetty.http.port=8080
```

Each line would be run as a jetty `start.jar` command with the added feature that we do property replacement, so that any properties that are on the line will replace the property within any generated `start.d/module.ini` file.

This keeps the jetty configuration in a single file, out of the pom.xml, yet protects against future releases changing modules.  We can even warn if the properties are no longer known!

thoughts?

Dang, looks like we will need another 9.4.0 RC which will push the release back a week, but could be worth it.

joakime:
How about just a `--save-props` command line that takes the properties provided on the command line and sets them in the appropriate `*.ini` file?

janbartel:
@joakime I like the idea of saving the props on the command line, but I don't really think its feasible to try and find out which .ini file they belong in: I think they'd have to be all put into a single.ini file that is always applied last called eg "override.ini" or something.

gregw:
@joakime what I'm proposing is to essentially have a `save-props` behaviour the default with any `--add-to-start=` on the command line.    Currently we don't specify properties when adding to start, so it will not break any existing usage.

But furthermore, the proposal for a `--configuration` option is there to allow such configuration to be put in a specific file under configuration control rather than being buried inside a pom.xml as part of a specific plugin configuration.

I see both features as separable, but both would be good to improve the configuration of jetty-bases in situations like we have here in jetty-runtime.     I'll experiment later today and confirm if they are viable.

@meltsufin I'll update this PR to the latest.  You can either approve now as correct behaviour or wait for the feature that will allow better separation of the configuration.

gregw:
See https://github.com/eclipse/jetty.project/issues/1139

Issue #72, opened 2016-11-21
Review test approaches


jmcc0nn3ll:
I have pushed up a couple of branches, jcm-test-split and jcm-test-combined.

Briefly, the idea is to support both local and remote style of testing, the local using a local docker setup and the remote were a docker image is created and run using flex.

By default, no tests are run.

> cd tests;
> mvn install -Plocal -Djetty.test.image=jetty:9.3

This will run local tests using a locally installed jetty:9.3 image as the FROM line for the docker file.

> cd tests;
> mvn install -Premote -Djetty.test.image=gcr.io/<project>/jetty9:9.3

This will run remote tests using the appropriate remote image via cloud builder and run the tests.

Both branches follow the same basic command line approach.  However the split one has both 'local' and 'remote' container directories.  The combined one I like a bit more since it follows the model that jetty uses for our session tests where we can have the same basic test for multiple session managers and just implement them where we need, in this simple example we have a Hello world test with both a local and remote implementation.

@meltsufin thoughts?

meltsufin:
@jmcc0nn3ll I like the combine approached more as well. There is a lot less code duplication that way.

jmcc0nn3ll:
ok, I'll drill down on that, clean up for checkstyle, and submit a PR for it

jmcc0nn3ll:
this has been resolved now

PR #80, opened 2016-11-29
test layout + initial sample smoke test for issue #72

jmcc0nn3ll:


tests/test-war-smoke/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM ${jetty.test.image}
+ADD root /app
```
meltsufin: I'm having an issue with this working. See: #85. I actually can't get the test to pass without adding `RUN chmod -R +r /app` to this Dockerfile.
Maybe we should stick with the WAR deployment without exploding it for now, because that's what we've always been supporting.

meltsufin: I'm using `jetty:9.3` that is built by `mvn clean install` from the project root. I think the problem is that the files in my working directory have different permissions and those permissions are propogated through Docker's `ADD`, but I haven't verified that yet. I'll dig a bit more.

jmcc0nn3ll: excellent, now it is making sense to me...  @gregw do you know where that link came from?

https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/jetty9/src/main/docker/Dockerfile#L41

jmcc0nn3ll: blame shows: 

https://github.com/GoogleCloudPlatform/jetty-runtime/commit/c367b6ccd98dc2c307adf3d43585f4322f1caa87

this intended usage?

meltsufin: I don't think that really explains it.
In any case, I was able to get the app to run with `ADD root /app`, but I still get a directory listing if I open `localhost:8080/` rather than `localhost:8080/hello`. So, there's still something not right with this approach. I really think we should switch to something like `ADD test-war-smoke-0.1.0-SNAPSHOT.war $JETTY_BASE/webapps/root.war`, and create a separate issue to make sure we have support for building images with exploded WAR.

jmcc0nn3ll: that would be because there is nothing in the war that maps to /, just the hello servlet on /hello

I'll push something on that soon

jmcc0nn3ll: Pushed, though should axe the commented out plugin in the test-war-smoke/pom.xml eventually, left as placeholder for unpack approach for time being.

going to / will just show a simple html file now, simple test is run against hello servlet which can all be expanded on quickly once the basic approach is deemed acceptable

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/HttpUrlUtil.java
```diff
@@ -0,0 +1,105 @@
...
+        http.setRequestProperty("User-Agent", "jetty-runtime/gcloud-testing-core(server-up)");
+        int statusCode = http.getResponseCode();
+        if (statusCode != HttpURLConnection.HTTP_OK) {
+          log.log(Level.FINER, "Waiting 2s for next attempt");
+          TimeUnit.SECONDS.sleep(2);
+        } else {
+          waiting = false;
+        }
+      } catch (MalformedURLException e) {
+        Assert.fail("Invalid URI: " + uri.toString());
```
meltsufin: Why not just remove this catch block or wrap it into an IllegalArgumentException? This will allow us to get rid of the junit-compile dependency and give more flexibility to the caller.

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/HttpUrlUtil.java
```diff
@@ -0,0 +1,105 @@
...
+   * @return the text of the response body
+   * @throws IOException if unable to get the text of the response body
+   */
+  public static String getResponseBody(HttpURLConnection http) throws IOException {
+    Charset responseEncoding = StandardCharsets.UTF_8;
+    if (http.getContentEncoding() != null) {
+      responseEncoding = Charset.forName(http.getContentEncoding());
+    }
+
+    try (InputStream in = http.getInputStream();
```
meltsufin: Can it be done with a one-liner with Apache Commons?
`String result = IOUtils.toString(http.getInputStream(), responseEncoding);`

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/ProcessUtil.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+
+/**
+ * Utilities for executing and collecting output from command line processes.
+ */
+public final class ProcessUtil {
```
meltsufin: This whole class is currently unused. Let's remove it until it's needed.

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/RemoteLog.java
```diff
@@ -0,0 +1,224 @@
...
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+public class RemoteLog {
```
meltsufin: Same here, it's not used as part of this PR. Let's remove it for now, and add it in as part of a separate PR that uses it.

tests/gcloud-testing-core/src/test/java/com/google/cloud/runtime/jetty/testing/RemoteLogTest.java
```diff
@@ -0,0 +1,23 @@
...
+import static org.hamcrest.Matchers.greaterThan;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.List;
+
+public class RemoteLogTest {
```
meltsufin: We don't need this if we remove the RemoteLog class.

tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/tests/HelloLocalIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Objects;
+
+public class HelloLocalIntegrationTest extends AbstractHelloIntegrationTest {
+  @BeforeClass
+  public static void before() throws Exception {
+    // give app time to start up
+    Thread.sleep(2000);
```
meltsufin: Can we use the `HttpUrlUtil.waitForServerUp` here instead?

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/AppDeployment.java
```diff
@@ -0,0 +1,68 @@
...
+
+    Objects.requireNonNull(projectId, "app.deploy.project");
+    Objects.requireNonNull(version, "app.deploy.version");
+
+    PROJECT_ID = projectId;
+    VERSION_ID = version;
+
+    String serviceId = null;
+
+    Path appYamlPath = MavenTestingUtils.getProjectFilePath("src/main/appengine/app.yaml");
```
joakime: (just providing some historical context)  The parsing of the `app.yaml` was to get the module/service name out of the active (in the maven reactor) project.  The passing in of `GAE_SERVICE` didn't exist at the time that the local integration testing steps execute (previously, that env variable only existed on the server side runtime, not the maven/surefire/junit runner side)

meltsufin: Understood, but is knowing the service name needed for local tests?

joakime: It was used to query the deployed application in remote testing scenarios. (both from an HTTP point of view, and also the various gcloud commands)

jmcc0nn3ll: @meltsufin basically we need to construct the following url, https://20161213-1806-dot-smoke-dot-jetty9-work.appspot.com

I seem to be getting a null property out of GCLOUD_PROJECT, are you sure that property exists within the plugin environment?  Do you know of another way to obtain that url?

tests/gcloud-testing-core/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+          <version>0.5.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud</artifactId>
+            <version>0.5.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty.toolchain</groupId>
+            <artifactId>jetty-test-helper</artifactId>
```
joakime: Note: jetty-test-helper has a dependency on junit and hamcrest.

jmcc0nn3ll:
I have most/all of the changes incorporated, just working on getting the remote testing working again and trying to remove the gcloud exec requirement for that project id...once I get remote working I'll take another pass through the pom's and update the pr

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/HttpUrlUtil.java
meltsufin: Sure, if it can do the same thing.

meltsufin:
Sounds good. Thanks!

tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/testing/AppDeployment.java
meltsufin: Disregard this whole comment. I realize that even for "remote tests" we need this info locally. Even though there are a number of other ways to get the app URL, such as parsing the gcloud CLI output or using the appengine-core-library, they're a bit involved. So, let's leave it to a separate PR.

meltsufin: Tracking in #93

jmcc0nn3ll:
@meltsufin pushed update that should address all the details above

meltsufin:
LGTM! :+1:

PR #89, opened 2016-12-08
Cleanup of jetty user and /app symlink

### LINKED ISSUE #85: /app not owned by jetty user
https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/jetty9/src/main/docker/Dockerfile#L41
It looks like with need to `chown -R jetty:jetty /app`, or am I missing something? @gregw @jmcc0nn3ll

gregw:
Fixes #85 by
- removed the symlink that is created in the Dockerfile
- updated docker-entrypoint.bash to look for content in app and only create the symlink if needed
- removed the setuid module
- add USER jetty to the Dockerfile

### DIFF
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -26,18 +26,17 @@ ENV GAE_IMAGE_LABEL ${docker.tag.long}
 # Create Jetty Home
 ENV JETTY_HOME ${jetty.home}
 ENV PATH $JETTY_HOME/bin:$PATH
-ENV TMPDIR /tmp/jetty
 ADD jetty-distribution.tar.gz $JETTY_HOME/..
-RUN mkdir $TMPDIR \
- && chown -R jetty:jetty $JETTY_HOME $TMPDIR
+RUN chown -R jetty:jetty $JETTY_HOME
 
 # Create Jetty Base
 ENV JETTY_BASE ${jetty.base}
+ENV TMPDIR /tmp/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
-RUN mkdir webapps \
+RUN mkdir webapps $TMPDIR \
  && java -jar $JETTY_HOME/start.jar --add-to-start=setuid \
- && chown -R jetty:jetty $JETTY_BASE
+ && chown -R jetty:jetty $JETTY_BASE $TMPDIR
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -24,24 +24,24 @@ ENV GAE_IMAGE_NAME jetty
 ENV GAE_IMAGE_LABEL ${docker.tag.long}
 
 # Create Jetty Home
-ENV JETTY_HOME /usr/local/jetty
+ENV JETTY_HOME ${jetty.home}
 ENV PATH $JETTY_HOME/bin:$PATH
 ENV TMPDIR /tmp/jetty
-ADD jetty-distribution.tar.gz /usr/local/
-RUN mv /usr/local/jetty-distribution-* $JETTY_HOME \
- && chown -R jetty:jetty $JETTY_HOME
+ADD jetty-distribution.tar.gz $JETTY_HOME/..
+RUN mkdir $TMPDIR \
+ && chown -R jetty:jetty $JETTY_HOME $TMPDIR
 
 # Create Jetty Base
-ENV JETTY_BASE /var/lib/jetty
+ENV JETTY_BASE ${jetty.base}
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
-RUN mkdir webapps /tmp/jetty \
+RUN mkdir webapps \
  && java -jar $JETTY_HOME/start.jar --add-to-start=setuid \
- && chown -R jetty:jetty /tmp/jetty $JETTY_BASE
+ && chown -R jetty:jetty $JETTY_BASE
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /
-RUN chmod +x /docker-entrypoint.bash
+RUN chmod +rx /docker-entrypoint.bash
 
 EXPOSE 8080
-CMD ["java","-Djetty.base=/var/lib/jetty","-jar","/usr/local/jetty/start.jar"]
+CMD ["java","-Djetty.base=${jetty.base}","-jar","${jetty.home}/start.jar"]
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -18,7 +18,9 @@ FROM ${docker.openjdk.image}
 RUN groupadd -r jetty \
  && useradd -r -g jetty jetty \
  && apt-get -q update \
- && apt-get -y -q --no-install-recommends install sudo 
+ && apt-get -y -q --no-install-recommends install sudo \
+ && apt-get clean \
+ && rm /var/lib/apt/lists/*_*
 
 # Create env vars to identify image
 ENV JETTY_VERSION ${jetty9.version}
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -35,14 +35,15 @@ RUN mv /usr/local/jetty-distribution-* $JETTY_HOME \
 ENV JETTY_BASE /var/lib/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
-RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
- && chown -R jetty:jetty $JETTY_BASE \
- && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
- && ln -s $JETTY_BASE/webapps/root /app
+RUN mkdir webapps /tmp/jetty \
+ && chown -R jetty:jetty /tmp/jetty $JETTY_BASE
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /
 RUN chmod +x /docker-entrypoint.bash
+
+# Use the Jetty USER
+USER jetty
 
 EXPOSE 8080
 CMD ["java","-Djetty.base=/var/lib/jetty","-jar","/usr/local/jetty/start.jar"]
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -16,7 +16,9 @@ FROM ${docker.openjdk.image}
 
 # create jetty group and user
 RUN groupadd -r jetty \
- && useradd -r -g jetty jetty
+ && useradd -r -g jetty jetty \
+ && apt-get -q update \
+ && apt-get -y -q --no-install-recommends install sudo 
 
 # Create env vars to identify image
 ENV JETTY_VERSION ${jetty9.version}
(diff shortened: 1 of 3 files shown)


meltsufin:
Thanks Greg! I just updated description to use issue [auto-close keyword](https://help.github.com/articles/closing-issues-via-commit-messages/), and created a more descriptive title.

gregw:
@meltsufin This is ready again for consideration.

Note that the openjdk image does an `apt-get update`, then installs some packages and then `apt-get clean`.  So I have followed that pattern here.    However I can see an argument to remove the clean from both images.

With the clean, every image has to update the apt repos before installing every module. This means they get a view of the repos at the time they are built rather than at the time the parent image was built.

However, it could be argued that when using a parent image, it would be best to install the packages that were available at the time the image was created.   If users want the latest/greatest, they are always free to update themselves.

Note that cleaning the cache does not reduce image size, as the files will still be added in one layer and removed in the next.  In fact an image that is extended several times in this way may have the apt cached added and deleted several times, so it may have several copies in the image file.

meltsufin:
I just came across these [docker best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/run) regarding apt-get, and they say that you don't actually need to run `apt-get clean` at all.

If you're removing the cache as part of the same `RUN` command, why would it be added to any layer?

gregw:
You are probably right that by being part of the same RUN command, it does not add to the layers.  However, shall we follow the best practises and remove the clean from openjdk and jetty runtimes?

joakime:
From that link ...

> In addition, cleaning up the apt cache and removing /var/lib/apt/lists helps keep the image size down. Since the RUN statement starts with apt-get update, the package cache will always be refreshed prior to apt-get install.

`apt-get clean` or `apt-get clean all` will remove the `/var/cache/apt` cache contents and update the internal metadata tracking of that cache.

`rm /var/lib/apt/lists/*_*` will remove the apt lists (different then the apt cache)

joakime:
Does the official google linux image have a bootstrap that cleans out the `apt cache` every time the docker image is started?

The ubuntu image does it like this : https://github.com/docker/docker/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105

gregw:
@joakime thanks for pointing out that text.  I think we are doing the right thing by `update&&install&&clean&&rm` in a single `RUN` command.  This will give us minimal size image and no cleanup to do on boot.

meltsufin:
Our [base image](https://github.com/GoogleCloudPlatform/debian-docker) is Debian. It looks like we do have the [auto clean setup](https://github.com/GoogleCloudPlatform/debian-docker/blob/master/third_party/docker/mkimage/debootstrap#L82). Doesn't this mean we can remove the `apt-get clean` statements without affecting the image size?

gregw:
Wow - I've been corrected back and forth 3 times in 5 minutes :) Yes that does indicate we can remove the cleans from our docker file in both jetty-runtime and openjdk-runtime.

joakime:
If you want to be careful of image size, then we should `apt-get clean` to produce a smaller image.

If we went the route of the debian `auto clean setup` then that means the image will contain the `/var/cache/apt` and everyone that downloads / uploads that image will have that directory, but the moment they run that image the `/var/cache/apt` is removed.

To explain this in more detail, lets use /jetty-runtime/ as an example.
And lets say the /openjdk-runtime/ was built without the `apt clean` step.

IIRC, The /jetty-runtime/ Dockerfile does no apt-get of its own, so ...

 1. Build /jetty-runtime/ Dockerfile
 2. FROM openjdk-runtime image (it has `/var/cache/apt` present, and downloads an image with `/var/cache/apt` present)
 3. Apply /jetty-runtime/ RUN commands
 4. openjdk-runtime image starts
 5. automatic debian 'auto clean setup' runs and removes `/var/cache/apt`
 6. RUN commands for /jetty-runtime/
 7. Stop /jetty-runtime/ image on specified new tag (does not contain `/var/cache/apt`)

This means that the /openjdk-runtime/ has the `/var/cache/apt`, but the resulting /jetty-runtime/ image does not.  Does this make sense?

joakime:
Ultimate the decision comes down to ...

1. Do we want a small image at runtime?
2. Do we want a small image for upload/download reasons?

If you want a small image at runtime, and don't care about upload/download sizes, then remove `apt-get clean`, as everyone will just use `FROM jetty-runtime` and then do their own thing, which will automatically clean up the results from the /jetty-runtime/ image.  Leaving the small image size question for the user of this image.

if you want a small image for upload/download of the /jetty-runtime/ base image reasons, then keep the `apt-get clean`

gregw:
Sorry guys I'm a bit slow this morning..., but now I think I'm fully up to speed:
- Jetty-runtime Dockerfile currently does no apt-get itself, but this PR adds one.
- the debian base image has been modified so that an `apt-get clean` will also do a `com/google/apphosting/runtime/jetty9`
- our openjdk image and this PR go 1 step further and also do `rm /var/lib/apt/lists/*_*`
- since the recommendation from Docker is to always do an update before any other installs, it makes little sense to leave a package cache in the image. 

So I think openjdk-runtime and this PR are doing the right thing.   My only question now is should the `rm /var/lib/apt/lists/*_*` be moved to the debootstrap?

joakime:
Note that you can see the layers of an image, and their individual sizes with ...

```
docker history <image-id>
```

Test it with, and without, the `apt-get clean` and see what the results are (instead of speculation based on debian behavior, docker behavior, and docker documentation interpretation)

gregw:
Testing this PR, the command to add sudo with a clean adds a 3.4MB layer to the image. Without the clean it add a 14.15MB layer to the image.  This is approx 3% of the image size and thus a saving I think is worth making.

meltsufin:
Ok, I think that settles it! :) Thanks @joakime and @gregw 
The note in Docker's guide is a bit misleading because it makes it sound like there's no point to running 'apt-get clean'.

meltsufin:
On a different note, I'm trying to run the image built from this PR and getting this:
```
$ docker run --rm -p8080:8080 jetty:9.4
/bin/bash: /docker-entrypoint.bash: Permission denied
```

meltsufin:
Sorry, going back to measuring the size of the layer with and without the clean, I'm not getting the same results as @gregw. For me, the layers and the resulting images are same size either way.

```
sha256:13fe747872d67c7105cc2b9d4541927475c2576f10f9886cf7117d33c3dad5e1   42 minutes ago       /bin/sh -c groupadd -r jetty  && useradd -r -g jetty jetty  && apt-get -q update  && apt-get -y -q --no-install-recommends install sudo  && apt-get clean  && rm /var/lib/apt/lists/*_*                                                                                                                                                          3.146 MB           


sha256:cead9d196f8d45df9edb797a5146b08ff834d7354c2b79cb080b9aeb9e0ed91f   4 minutes ago       /bin/sh -c groupadd -r jetty  && useradd -r -g jetty jetty  && apt-get -q update  && apt-get -y -q --no-install-recommends install sudo  && rm /var/lib/apt/lists/*_*                                                                                                                                                                            3.146 MB     
```

gregw:
@meltsufin Ah, I also removed the `rm /var/lib/apt/lists/*_*`, so I guess that is where the savings are!

On the image generated by this PR I can do `docker run --rm -it jetty:9.4 bash` without permission issues.  Without the bash, it still runs and fails with the 'No Context' ISE as expected.

gregw:
I can confirm that removing the clean makes no difference to the image size, it is rm of the apt lists that saves 10MB.

I also notice the following warning when adding sudo:
```
Get:1 http://httpredir.debian.org/debian/ jessie/main sudo amd64 1.8.10p3-1+deb8u3 [855 kB]
debconf: delaying package configuration, since apt-utils is not installed
```

Should the base image have apt-utils installed?  see https://github.com/phusion/baseimage-docker/issues/319

meltsufin:
I'd rather not add apt-utils if it's not strictly necessary and will only be taking up more space.

I'm still getting 'Permission denied' and it seems to be happening on this line.
```shell
# Switch user to jetty
if [ "$UID" = 0 ] ; then
  exec sudo -HEu jetty -- "$0" "$@"
fi
```
Are you sure your image was fully rebuilt? Can you try this command sequence to verify?
```
$ docker run --rm -it --entrypoint=bash jetty:9.4
root@62a10b3732fb:/var/lib/jetty# cd /
root@62a10b3732fb:/# grep 'exec sudo' docker-entrypoint.bash 
  exec sudo -HEu jetty -- "$0" "$@"
root@62a10b3732fb:/# ./docker-entrypoint.bash 
/bin/bash: ./docker-entrypoint.bash: Permission denied
```

gregw:
@meltsufin WFM??

```
[2224] docker run --rm -it --entrypoint=bash jetty:9.4
root@9ac8cc85d6aa:/var/lib/jetty# cd /
root@9ac8cc85d6aa:/# grep 'exec sudo' docker-entrypoint.bash
  exec sudo -HEu jetty -- "$0" "$@"
root@9ac8cc85d6aa:/# id
uid=0(root) gid=0(root) groups=0(root)
root@9ac8cc85d6aa:/# ./docker-entrypoint.bash
-XX:InitialHeapSize=9642704896 -XX:MaxHeapSize=9642704896 -XX:+ParallelRefProcEnabled -XX:+PrintCommandLineFlags -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseG1GC 
openjdk version "1.8.0_111"
OpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-2~bpo8+1-b14)
OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)

2016-12-13 22:51:00.195:INFO::main: Logging initialized @497ms to org.eclipse.jetty.util.log.StdErrLog
2016-12-13 22:51:00.447:INFO:oejs.Server:main: jetty-9.4.0.v20161208
2016-12-13 22:51:00.485:INFO:oejdp.ScanningAppProvider:main: Deployment monitor [file:///var/lib/jetty/webapps/] at interval 0
2016-12-13 22:51:00.504:INFO:oejs.AbstractConnector:main: Started ServerConnector@6d00a15d{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
2016-12-13 22:51:00.504:INFO:oejs.Server:main: Started @806ms
org.eclipse.jetty.server.Server@614c5515 - STARTED
...
```

what user are you before you try to run the bash script?

meltsufin:
@gregw I think I know what the problem is. It's the permissions on the `/docker-entrypoint.bash`.
```
root@d9bd4da47763:/# ls -l docker-entrypoint.bash 
-rwxr-x--x 1 root root 1395 Dec 14 15:55 docker-entrypoint.bash
```
Since we're now reading and executing the script under the jetty user, it needs to be readable by jetty.
The fix is as simple as changing the chmod line in the Dockerfile to:
```
RUN chmod +rx /docker-entrypoint.bash
```

jetty9/src/main/docker/Dockerfile
```diff
@@ -16,7 +16,11 @@ FROM ${docker.openjdk.image}
 
 # create jetty group and user
 RUN groupadd -r jetty \
- && useradd -r -g jetty jetty
+ && useradd -r -g jetty jetty \
+ && apt-get -q update \
+ && apt-get -y -q --no-install-recommends install sudo \
```
meltsufin: Note that [Docker official best practices doc](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/user) recommends against installing sudo. I think we should consider alternatives.

meltsufin:
All other language runtimes execute under root, and it seems to be acceptable to do that. We can simplify things quite a bit by switching to running jetty under root.

gregw:
@meltsufin I think in order to simplify the process, I should update this PR to just keep using the setuid mechanism and then we can discuss the pros/cons of using the jetty user (or not) in #99

gregw:
@meltsufin would be good to get these changes in to simplify other PRs in progress

PR #91, opened 2016-12-09
throw error on startup if one occurs

jmcc0nn3ll:
@meltsufin This _should_ fail the startup of the server if the WebAppContext has issues on startup.

### DIFF
--- a/jetty9-base/src/main/java/com/google/cloud/runtimes/jetty9/DeploymentCheck.java
+++ b/jetty9-base/src/main/java/com/google/cloud/runtimes/jetty9/DeploymentCheck.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.runtimes.jetty9;
+
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.component.LifeCycle;
+
+import java.util.Arrays;
+
+/**
+ * Checks whether the application deployed normally.
+ */
+public class DeploymentCheck extends AbstractLifeCycle.AbstractLifeCycleListener {
+  @Override
+  public void lifeCycleStarted(LifeCycle bean) {
+    if (bean instanceof Server) {
+      Server server = (Server)bean;
+      Connector[] connectors = server.getConnectors();
+      if (connectors == null || connectors.length == 0) {
+        server.dumpStdErr();
+        throw new IllegalStateException("No Connector");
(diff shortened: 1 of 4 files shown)


meltsufin:
Thanks, Jesse! I'll test it out next week.

gregw:
@jmcc0nn3ll Your patch does indeed allow an init exception to be thrown externally to the context.  However because this context is being started by the deployer an not from a direct embedded usage, the exception does not cause the server to exit.       I'll have a look at how we can configure the deployer to achieve that.... stand by....

gregw:
I've updated this PR with a DeploymentCheck class that is attached to the server life cycle.  Once the server is started, it checks for a started Connector and Context, if either is not found then an ISE is thrown which does stop the server...  it also has revealed the first bug in 9.4.0 that we have a NPE when reporting that ISE, which is a cut and paste problem that will be fixed in 9.4.1 - for now it just produced an unnecessary extra exception as the server exits.

jetty9-base/src/main/java/com/google/cloud/runtimes/jetty9/DeploymentCheck.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.component.LifeCycle;
+
+public class DeploymentCheck extends AbstractLifeCycle.AbstractLifeCycleListener {
+  @Override
+  public void lifeCycleStarted(LifeCycle bean) {
+    if (bean instanceof Server) {
+      Server server = (Server)bean;
+      Connector[] connectors = server.getConnectors();
+      if (connectors.length == 0 || !connectors[0].isStarted()) {
```
meltsufin: Will there always be just one connector, or should we loop through them all?

jetty9-base/src/main/jetty-base/etc/gae-web.xml
```diff
@@ -2,6 +2,9 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_3.dtd">
 <Configure class="org.eclipse.jetty.webapp.WebAppContext">
 
+  <!-- Throw Exception on Start -->
```
meltsufin: This comment is a little confusing. Can we put something more descriptive like "Propagate application startup exceptions to server", or whatever makes sense.

gregw:
Several reasons why deployer does not propagate errors by default:  some users simply do not want it;  for hot deploy there is nowhere to propagate to.

However, we will look at making start errors able to be propagated.

jetty9-base/src/main/java/com/google/cloud/runtimes/jetty9/DeploymentCheck.java
gregw: There can be more connectors.   I was just testing minimal conditions of at least 1 working connector, which I guess may not be the first.    So I can test all, but do we check for all started, or just 1 of them started?

meltsufin: I guess, all of them? What are the other connectors that we will normally have?

gregw: Normally we only have 1.  But a user may extend to have other connectors for private or other purposes.  I'll check that all connectors are started.

gregw:
@meltsufin I can do for async-support.

Note that the CLA checker is confused again and It's saying I can't yet merge?

gregw:
@meltsufin do you want this for https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/async-support or https://github.com/GoogleCloudPlatform/jetty-gae-compat-runtime?

I could just copy the class into async-cupport, or I could consume a maven artefact produced by jetty-runtime?

meltsufin:
@gregw Can you just copy the class and config to https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/async-support?

I'll try to fix the CLA checks.

meltsufin:
Sorry @gregw had to squash your commit into Jesse's to make this work. :)

PR #94, opened 2016-12-14
add basic jsp jstl environment test

jmcc0nn3ll:
taking the test setup for a spin, here is a basic jsp/jstl environment test for local and remote

### DIFF
--- a/tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/tests/JspRemoteIntegrationTest.java
+++ b/tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/tests/JspRemoteIntegrationTest.java
@@ -1,0 +1,25 @@
+package com.google.cloud.runtimes.jetty.tests;
+
+
+import com.google.cloud.runtime.jetty.testing.AppDeployment;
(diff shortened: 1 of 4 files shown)


tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/tests/JspRemoteIntegrationTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+public class JspRemoteIntegrationTest extends AbstractJspIntegrationTest {
+
+  @BeforeClass
+  public static void isServerUp() {
+    HttpUrlUtil.waitForServerUp(AppDeployment.SERVER_URI, 5, TimeUnit.MINUTES);
```
meltsufin: Testing for server up seems to be a common theme across all of these tests. Maybe we can pull this up into a parent class like `AbstractIntegrationTest`. Also, AppDeployment.SERVER_URI should probably be made to default to localhost, so that we can use a single `isServerUp()` for both local and remote tests.

jmcc0nn3ll:
sure, I have a bit of work on that now but nothing in a form to into PR yet

PR #99, opened 2016-12-16
Run jetty as root

meltsufin:
This PR is partly based on #89, but makes jetty execute as root instead of the 'jetty' user.


jmcc0nn3ll:
I think the best policy is to run the server instance as a non-root user, however how much of that is seeped in tradition and ameliorated by the runtime architecture itself is open to debate.

jetty9/src/main/docker/Dockerfile
```diff
@@ -14,35 +14,28 @@
...
 # Create Jetty Home
-ENV JETTY_HOME /usr/local/jetty
+ENV JETTY_HOME ${jetty.home}
 ENV PATH $JETTY_HOME/bin:$PATH
 ENV TMPDIR /tmp/jetty
-ADD jetty-distribution.tar.gz /usr/local/
-RUN mv /usr/local/jetty-distribution-* $JETTY_HOME \
- && chown -R jetty:jetty $JETTY_HOME
+RUN mkdir $TMPDIR
+ADD jetty-distribution.tar.gz $JETTY_HOME/..
```
meltsufin: @jmcc0nn3ll is there any way to do some Maven magic to have `jetty-distribution.tar.gz` not have all the files in a subdirectory `jetty-distribution-${jetty9.version}` but rather in the root?

jmcc0nn3ll: First off, the jetty-home artifact might be better to use, we started using it in Jetty-9.4 and it is basically the distribution minus the test applications, etc...bit of a smaller footprint.

Secondly, that extra directory is an artifact of it being an archive and no one likes to unpack a zip file and have a mess of files show up in their current directory.  I have tinkered around in the past and never found a clean way of doing what you are asking through Maven.  Short of some antrun after unpack...

  ```
 <copy todir="target/docbkx/html/documentation/9.3.9.v20160517">
    <fileset dir="target/documentation-resource/9.3.9.v20160517/jetty-documentation-9.3.9.v20160517/html"/>
</copy>
```

jmcc0nn3ll: For the tarball:

jetty-home: 8005781
jetty-distribution: 17763409

jmcc0nn3ll: I can't promise it is 100% drop in replacement right now, it should be close and if it isn't then likely something we need to tweak because it is ultimately meant to be used for these sorts of things.

meltsufin: Sounds like a good idea. Tracking #102

gregw:
This PR has some good cleanup changes not related to running as root vs jetty, so I've added them to PR #89 for #85 as it has some general cleanup already.   I'll next update this branch to be just the change to run root and we can discuss that.

gregw:
There are definitely benefits of running as root:
- Simpler Dockerfiles with no need to chown or chmod files as they are added.
- Simpler to extend the image.
- able to run jetty as pure java with not setuid module.

However there are some problems as well:
- It is different to the the behaviour of the standard docker jetty image.
- It is against the [docker best practises](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#user)
- Less detection of incorrectly configured applications that may be trying to access incorrect files and/or resources.
- It removes a layer of security and thus increases the expose of webapplications deployed on the jetty-runtime.   I admit the chances of any significant security breach are pretty slim, however I have very little trust in the quality of most complex webapplications written and would expect many of them to contain security problems that will allow the hosts file system to be written to.  If running as root, then any such webapp deployed would essentially give any remote user root access to the image.   While this in itself is not a significant impact on Google infrastructure, it may allow deeper penetration of the attacked webapp than otherwise may have been possible.    

So on balance, the benefits of running as root are pretty minor, but the downside, while unlikely is potentially very severe.    Thus I can't recommend that the jetty-runtime should run as root, but will happily accept the benefits if directed to do so. 

Note that [OpenShift is changing](http://blog.dscpl.com.au/2015/12/don-run-as-root-inside-of-docker.html) to prohibit root docker images from being deployed.

gregw:
I've updated this branch to https://github.com/GoogleCloudPlatform/jetty-runtime/tree/as_root-85 which is the as root changes made against the latest  https://github.com/GoogleCloudPlatform/jetty-runtime/tree/master-85, which itself has been merged with the latest master.

jetty9/src/main/docker/Dockerfile
gregw: I'm going to comment on the version extension question in #102

meltsufin:
@gregw I hear your points about running as jetty user being more secure, but I'm not fully convinced that it is dangerous specifically in our scenario where jetty is essentially the only process that runs in the container and already has access to all inputs and outputs of the application. I do like the additional benefit of potentially preventing user error, but it can also work the other way blocking user apps from doing something that is reasonable.
Also, note that the java process in the openjdk-image is running under root, and I've also confirmed that all other GCP official images run the apps under root.

In any case, I'm not rushing to merge this in. I think we should discuss the pros and cons a bit more.

meltsufin:
Closing because currently there are no immediate benefits for the switch to root, while introducing deviation from Docker best-practices.

PR #100, opened 2016-12-16
Turn on local tests by default

meltsufin:
I'm exploring the option of having local tests run by default, since they run so quickly.
One concern I have is the statement we have in the design doc saying "local and remote testing are mutually exclusive". I'm not sure if this has to be the case. @jmcc0nn3ll

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
     <docker.tag.prefix></docker.tag.prefix>
     <docker.tag.short>${docker.tag.prefix}9.${jetty9.minor.version}</docker.tag.short>
     <docker.tag.long>${docker.tag.prefix}9.${jetty9.minor.version}-${maven.build.timestamp}</docker.tag.long>
-    <docker.openjdk.image>gcr.io/google_appengine/openjdk:8-2016-12-08_19_56</docker.openjdk.image>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -286,6 +286,9 @@
   <profiles>
     <profile>
       <id>tests</id>
+      <activation>
```
jmcc0nn3ll: I think if you want to make tests run by default then we ought to do away with the tests profile and just add it to the modules list.

meltsufin: I was thinking about that too, but having it in a profile also allows you to quickly skip it with `-P!tests`.

meltsufin:
Local tests can use the `FROM gcr.io/...` as well as remote ones.

jmcc0nn3ll:
@meltsufin Since we seem to have settled on mutually exclusive are you still keen on local enabled by default?

With a bit profile activation magic we could set up local tests to run by default and if a -Dtest.remote property were used it could deactivate local testing off and activate the remote testing profile.  No more -Plocal and -Premote, just adding of -Dtest.remote property if you want to do remote testing.

Even if tests are not run by default, there is some value in the tests being part of the base build (making artifacts but not executing tests).  Local test running suppression could be achieved with  -P!test.local or something akin to that.

meltsufin:
I was thinking exactly the same. You'll see when I submit the review in a sec.

meltsufin:
Incorporated as part of #104

PR #107, opened 2016-12-21
Issue #88 disallow directory listing

### LINKED ISSUE #88: When app is missing a welcome file, directory listing is shown
If the app doesn't have a file like index.jsp in the web root, a directory listing is shown.

```
Directory: /

META-INF/ 	4096 bytes 	Dec 7, 2016 10:42:46 AM
WEB-INF/ 	4096 bytes 	Dec 7, 2016 10:42:02 AM
```

This seems to be the default behavior of Jetty. Is this something we want to keep, or should we disable it if possible?

gregw:
Fixes #88 by disallowing directory listings


jetty9-base/src/main/jetty-base/etc/gae-override-web.xml
```diff
@@ -0,0 +1,26 @@
...
+<!-- 
+  This web.xml format file is an override file that is applied to the webapp AFTER
+  it has been configured by the default descriptor and the WEB-INF/web.xml descriptor.
+  It can be used to override init-params and/or add/change additional servlets, filters
+  and mappings.
+-->
+
+  <servlet>
+    <servlet-name>default</servlet-name>
+    <init-param>
```
meltsufin: I think we need a comment here specifying what it's actually doing -- disallowing directory listing when there is no welcome file.

gregw:
@meltsufin added extra comments and a test

PR #113, opened 2016-12-28
Remove Circle CI and fix Travis CI configuration

meltsufin:
Enabling local tests on Travis CI.
Remote testing on Travis will be in a different PR.


.travis.yml
```diff
@@ -18,14 +18,12 @@ language: java
 
 services:
   - docker
-  
+
 jdk:
   - oraclejdk8
-install: mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
aslo: You should definitely override the `install` step here, perhaps with either a no-op, or something like `mvn dependency:resolve` which is what we do in the openjdk-runtime. By default, travis runs `mvn install` for the install phase of the build, which I believe will result in running the same tests twice (since we are running `mvn verify` in the `script` phase). 

https://docs.travis-ci.com/user/languages/java/#Dependency-Management

meltsufin: `mvn dependency:resolve` actually fails here. I'll put a noop for install for now.

meltsufin: Cut 30 seconds off the build. Thanks!

PR #115, opened 2016-12-29
Add structure tests

aslo:
An Initial stab at it. @meltsufin Let me know if you think anything major is missing.


aslo:
@meltsufin I created another issue to track those, so we can get these tests merged into the build process sooner.  see https://github.com/GoogleCloudPlatform/jetty-runtime/issues/116

PR #117, opened 2017-01-03
Resolve #109 local and remote testing works, -Ptest.remote,test.local

### LINKED ISSUE #109: Allow execution of both local and remote tests as part of one build

jmcc0nn3ll:
local is default
remote can be activated with -Ptest.remote
both with -Ptest.remote,test.local

figured out part of what is bothersome, the appengine-maven-plugin seems to force a complete new lifecycle so most all the plugins run twice

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,16 +61,13 @@ script:
   # execute remote tests
   - |
     if $DO_REMOTE; then
-      # deploy image to gcr
-      image=gcr.io/travis-app-maven-plugin/jetty_test:`date +%s`
-      docker tag jetty:9.4 $image
       # do the remote test
-      gcloud docker -q -- push $image && mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean -Djetty.test.image=$image -B -q
+      mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean -B -q
     fi
 
-after_script:
-  - |
-    if $DO_REMOTE; then
-      # cleanup deployed image
-      gcloud beta container images delete gcr.io/travis-app-maven-plugin/jetty_test@ -q
-    fi
+#after_script:
+#  - |
+#    if $DO_REMOTE; then
+#      # cleanup deployed image
+#      gcloud beta container images delete gcr.io/travis-app-maven-plugin/jetty_test@ -q
+#    fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ script:
       image=gcr.io/travis-app-maven-plugin/jetty_test:`date +%s`
       docker tag jetty:9.4 $image
       # do the remote test
-      gcloud docker -q -- push $image && mvn clean verify -Ptest.remote,test.local -Djetty.test.image=$image -B -q
+      gcloud docker -q -- push $image && mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean -Djetty.test.image=$image -B -q
     fi
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
       image=gcr.io/travis-app-maven-plugin/jetty:`date +%s`
       docker tag jetty:9.4 $image
       # do the remote test
-      gcloud docker -q -- push $image && mvn clean verify -Ptest.remote -Djetty.test.image=$image -B -q
+      gcloud docker -q -- push $image && mvn clean verify -Ptest.remote,test.local -Djetty.test.image=$image -B -q
     fi
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,7 @@ install:
 
 script:
   # execute regular local build
-  - mvn clean verify
-  #- mvn clean verify -B -V -q
+  - mvn clean verify -B -V -q
   # execute remote tests
   - |
     if $DO_REMOTE; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,13 +61,6 @@ script:
   # execute remote tests
   - |
     if $DO_REMOTE; then
-      # do the remote test
+      # do the remote test, will use default image setting and should clean up remote test image resources
       mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean -B -q
     fi
-
-#after_script:
-#  - |
-#    if $DO_REMOTE; then
-#      # cleanup deployed image
-#      gcloud beta container images delete gcr.io/travis-app-maven-plugin/jetty_test@ -q
-#    fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,8 @@ install:
 
 script:
   # execute regular local build
-  - mvn clean verify -B -V -q
+  - mvn clean verify
+  #- mvn clean verify -B -V -q
   # execute remote tests
   - |
     if $DO_REMOTE; then
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -30,7 +30,7 @@
   <properties>
     <checkstyle.skip>false</checkstyle.skip>
     <!-- default image name for local tests -->
-    <jetty.test.image>jetty:${docker.tag.short}</jetty.test.image>
+    <jetty.test.image>jetty:${docker.tag.long}</jetty.test.image>
   </properties>
   <modules>
     <module>gcloud-testing-core</module>
@@ -91,16 +91,34 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.19.1</version>
           <configuration>
             <excludes>
               <exclude>**/*IntegrationTest.java</exclude>
             </excludes>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>io.fabric8</groupId>
+          <artifactId>docker-maven-plugin</artifactId>
+          <version>0.17.2</version>
+          <configuration>
+            <images>
+              <image>
+                <name>${project.artifactId}:${project.version}</name>
+                <alias>${project.artifactId}</alias>
+                <run>
+                  <ports>
+                    <port>${local.app.port}:8080</port>
+                  </ports>
+                </run>
+              </image>
+            </images>
+          </configuration>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
+      <!-- ensure no artifacts are published during deploy -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-deploy-plugin</artifactId>
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -40,16 +40,6 @@
     <pluginManagement>
       <plugins>
         <plugin>
-          <groupId>com.google.cloud.tools</groupId>
-          <artifactId>appengine-maven-plugin</artifactId>
-          <configuration>
-            <deploy.project>${app.deploy.project}</deploy.project>
-            <deploy.version>${maven.build.timestamp}</deploy.version>
-            <deploy.promote>false</deploy.promote>
-            <dockerDirectory>${project.build.directory}/docker</dockerDirectory>
-          </configuration>
-        </plugin>
-        <plugin>
           <groupId>com.spotify</groupId>
           <artifactId>docker-maven-plugin</artifactId>
           <configuration>
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -40,24 +40,6 @@
     <pluginManagement>
       <plugins>
         <plugin>
-          <groupId>com.spotify</groupId>
-          <artifactId>docker-maven-plugin</artifactId>
-          <configuration>
-            <imageName>${project.artifactId}</imageName>
-            <imageTags>
-              <imageTag>${project.version}</imageTag>
-            </imageTags>
-            <dockerDirectory>${project.build.directory}/docker</dockerDirectory>
-            <resources>
-              <resource>
-                <targetPath>/</targetPath>
-                <directory>${project.build.directory}</directory>
-                <include>${project.build.finalName}.war</include>
-              </resource>
-            </resources>
-          </configuration>
-        </plugin>
-        <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-failsafe-plugin</artifactId>
           <version>2.19.1</version>
@@ -90,7 +72,7 @@
         <plugin>
           <groupId>io.fabric8</groupId>
           <artifactId>docker-maven-plugin</artifactId>
-          <version>0.17.2</version>
+          <version>0.19.0</version>
           <configuration>
             <images>
               <image>
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -129,6 +129,11 @@
         <targetPath>${project.build.directory}/appengine-staging</targetPath>
       </resource>
       <resource>
+        <directory>src/main/docker</directory>
+        <filtering>true</filtering>
+        <targetPath>${project.build.directory}/appengine-staging</targetPath>
+      </resource>
+      <resource>
         <directory>src/main/resources</directory>
         <filtering>true</filtering>
         <targetPath>
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -29,6 +29,7 @@
   <packaging>pom</packaging>
   <properties>
     <checkstyle.skip>false</checkstyle.skip>
+    <!-- default image name for local tests -->
     <jetty.test.image>jetty:${docker.tag.short}</jetty.test.image>
   </properties>
   <modules>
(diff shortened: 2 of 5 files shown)


tests/test-war-smoke/pom.xml
```diff
@@ -251,12 +228,46 @@
...
-                  <deploy.version>${maven.build.timestamp}</deploy.version>
-                  <deploy.promote>false</deploy.promote>
-                  <dockerDirectory>${project.build.directory}/docker
-                  </dockerDirectory>
+                  <executable>gcloud</executable>
+                  <arguments>
+                    <argument>docker</argument>
+                    <argument>-q</argument>
+                    <argument>--</argument>
+                    <argument>push</argument>
```
meltsufin: After the tests complete, it would be nice to cleanup the deployed test image from GCR, similarly how I did in Travis config:
`gcloud docker -q -- rmi -f $image`

tests/pom.xml
```diff
@@ -54,10 +83,32 @@
...
+              <exclude>**/*IntegrationTest.java</exclude>
+            </excludes>
           </configuration>
         </plugin>
       </plugins>
     </pluginManagement>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
```
meltsufin: What's the purpose of configuring the maven-deploy-plugin here?

tests/test-war-smoke/pom.xml
```diff
@@ -51,69 +52,37 @@
...
-          <value>local</value>
-        </property>
       </activation>
       <build>
         <plugins>
           <plugin>
-            <groupId>com.spotify</groupId>
-            <artifactId>docker-maven-plugin</artifactId>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-failsafe-plugin</artifactId>
```
meltsufin: Can the maven-failsafe-plugin config be pulled up to the parent test pom? There doesn't seem to be anything module-specific here.

jmcc0nn3ll: I'll do a pass through on all these now

tests/test-war-smoke/pom.xml
```diff
@@ -152,12 +121,13 @@
...
-          <name>test.mode</name>
-          <value>remote</value>
-        </property>
-      </activation>
+      <properties>
+        <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
+        <gcloud-projectId-file>${project.build.directory}/gcloud-projectid.properties
+        </gcloud-projectId-file>
+        <!-- remote tests require image in gcr.io -->
+        <jetty.test.image>gcr.io/${app.deploy.project}/jetty:${docker.tag.short}</jetty.test.image>
```
jmcc0nn3ll: kk

tests/test-war-smoke/pom.xml
jmcc0nn3ll: exec plugin would run at parent and each other step as well, to make a new module you'll copy this pom over and tweak anyway so not a huge burden I wouldn't think

some more of the config can probably be pulled up into plugin management, i did some before

tests/pom.xml
jmcc0nn3ll: keeps the war/docker file from ever being deployed into oss which is configured in the tlp

tests/test-war-smoke/pom.xml
meltsufin: Does `<inherited>false</inherited>` on the plugin help here at all to avoid re-running it in the child modules?

tests/pom.xml
meltsufin: OK, can you add a code comment here to point that out?

tests/test-war-smoke/pom.xml
jmcc0nn3ll: @meltsufin do you know there are two different usages of maven.build.timestamp.format?  the TLP has it as yyyy-MM-dd_HH_mm but the one that passes it into the appengine deploy in tests is yyyyMMdd-HHmm

jmcc0nn3ll: ah, gcloud restriction

GCLOUD: ERROR: (gcloud.app.deploy) argument --version/-v: Bad value [2017-01-03_19_43]: May only contain lowercase letters, digits, and hyphens. Must begin and end with a letter or digit. Must not exceed 63 characters.

jmcc0nn3ll: This is problematic because the tag carries the ${maven.build.timestamp} along with it and the change in the ${maven.build.timestamp.format} between pom.xml files changes the value of the ${docker.tag.long} property.

tests/test-war-smoke/pom.xml
jmcc0nn3ll: nope, that should prevent configuration making it into child pom's but I don't believe that applies for <executions/>

What would work here is matching the id of the exec plugin execution and then unbind the <phase>..it would match but not execute.  But...the issue there is that if you want it to push  in the parent and unbind in the child then you only ever push when running the parent, so you can't run just the build in the child and have image pushed.  Not a huge deal but if you want to see it work like that I'll just add a comment to that affect.

jmcc0nn3ll: i added an execution for this

jmcc0nn3ll: Note that if only push in the parent as mentioned in the previous review then having this in the child node would remove it for other artifacts.  I have it bound to post-integration-test atm but clean might be better.

tests/test-war-smoke/pom.xml
```diff
@@ -152,18 +93,22 @@
...
+      <properties>
+        <!-- change the build timestamp to something palatable for gcloud -->
+        <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
+        <gcloud-projectId-file>${project.build.directory}/gcloud-projectid.properties
+        </gcloud-projectId-file>
+        <!--
+           ~ remote tests require image in gcr.io
+           ~ if local tests are enabled with remote, this image is used locally as well
+           -->
+        <jetty.test.image>gcr.io/${app.deploy.project}/jetty:${docker.tag.short}</jetty.test.image>
```
meltsufin: I would use the `${docker.tag.long}` here to avoid possible collisions.

tests/pom.xml
```diff
@@ -39,6 +40,34 @@
     <pluginManagement>
       <plugins>
         <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>appengine-maven-plugin</artifactId>
+          <configuration>
+            <deploy.project>${app.deploy.project}</deploy.project>
+            <deploy.version>${maven.build.timestamp}</deploy.version>
```
meltsufin: I don't think you need to specify `deploy.project` and `deploy.version` since you're using system properties `app.deploy.project` and `app.deploy.version` below.

tests/test-war-smoke/pom.xml
```diff
@@ -228,19 +154,26 @@
...
+                <goals>
+                  <goal>integration-test</goal>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <skipTests>false</skipTests>
+                  <systemPropertyVariables>
+                    <test.mode>remote</test.mode>
+                    <app.deploy.project>${app.deploy.project}</app.deploy.project>
+                    <app.deploy.version>${maven.build.timestamp}</app.deploy.version>
```
meltsufin: For the sake of consistency, can we use `${docker.tag.long}` here for the version?

tests/test-war-smoke/pom.xml
```diff
@@ -251,12 +184,84 @@
...
+                  <goal>exec</goal>
+                </goals>
+                <configuration>
+                  <executable>gcloud</executable>
+                  <arguments>
+                    <argument>--quiet</argument>
+                    <argument>app</argument>
+                    <argument>versions</argument>
+                    <argument>delete</argument>
+                    <argument>${maven.build.timestamp}</argument>
```
meltsufin: you should be able to use `${app.deploy.version}` here.

tests/test-war-smoke/pom.xml
```diff
@@ -251,12 +184,84 @@
...
-                  <deploy.version>${maven.build.timestamp}</deploy.version>
-                  <deploy.promote>false</deploy.promote>
-                  <dockerDirectory>${project.build.directory}/docker
-                  </dockerDirectory>
+                  <executable>gcloud</executable>
+                  <arguments>
+                    <argument>docker</argument>
+                    <argument>-q</argument>
+                    <argument>--</argument>
+                    <argument>rmi</argument>
```
meltsufin: This whole command doesn't actually work. See #118 for the fix.

Also, I'm concerned about the whole image cleanup from GCR in the case where the user specified `-Djetty.test.image=....` Either the delete will fail, or we'll delete an image that we didn't push.

tests/test-war-smoke/pom.xml
jmcc0nn3ll: There was an issue here because the property for docker.tag.long was based upon the value of the maven.build.timestamp variable at the time of interpolation...and that changes between the build in jetty9 and the test cases because of the change in maven.build.timestamp.format.

tests/test-war-smoke/pom.xml
jmcc0nn3ll: Agreed, this is problematic in trying to take care of this sort of thing in a maven build.  Maven is very much a convention over configuration system with an emphasis on reproducible builds. These sorts of things are hard to reconcile...unless we encode the behavior in a maven plugin that can programmatically decide to do things like this...we are sort of at the mercy of doing it or not.  Safest thing to do would be to put it in a profile that you would have to add into the mix to clean up remote images like that.  Any way you look at it, it is awkward.  Trying to tie the execution of a plugin like this to some conditions outside of the plugin itself (ie the value of the jetty.test.image variable and if another plugin happened to have deployed a new image or not)...not simple.

meltsufin: I like the idea of an additional profile that deals with this. The profile would be activated when `jetty.test.image` is not specified and remote tests are on. This profile would push the image before tests and delete it after the tests. If it's not possible to automatically activate the profile, we can just require it to be explicitly turned on. WDYT?

jmcc0nn3ll: i am playing around with this, have it sort of working but will try and sort out the long tag as part of this as well

jmcc0nn3ll: can't really activate on two conditions sadly, i have it working until try and just run local tests and try and use default test image, the test deploy still triggers and then just fails trying to deploy to docker.io :/

meltsufin: How about just requiring the explicit activation of the profile that deploys image to GCR?

jmcc0nn3ll: that works, still can't work out a way to use docker.tag.long in this context though..I do have the separate profile working locally

meltsufin: We can drop the `docker.tag.long` thing. We can deal with that issue separately and potentially in a different way.

jmcc0nn3ll: pushed, added a test.remote.deploy and test.remote.clean profile for pushing and cleaning these remote images

jmcc0nn3ll:
think most everything is addressed with latest push, updated docs and the following should work generally

> mvn -Ptest.local,test.remote,test.remote.deploy,test.remote.clean install

This will run local tests, deploy a updated docker container, run the remote tests based on that container and clean up the remote docker container.  When test.remote profile is active without the jetty.test.image property a default value will be used.

tests/test-war-smoke/pom.xml
```diff
@@ -228,35 +167,146 @@
...
-                  <deploy.promote>false</deploy.promote>
-                  <dockerDirectory>${project.build.directory}/docker
-                  </dockerDirectory>
+                  <executable>gcloud</executable>
+                  <arguments>
+                    <argument>beta</argument>
+                    <argument>container</argument>
+                    <argument>images</argument>
+                    <argument>delete</argument>
+                    <argument>${jetty.test.image}</argument>
```
meltsufin: This whole command doesn't actually work, it turns out. We need to specify the sha256 digest of the image instead of the version tag. See: https://cloud.google.com/sdk/gcloud/reference/beta/container/images/delete
Do you think it's possible to capture the image digest in a maven property somehow? It would be very helpful.

tests/test-war-smoke/pom.xml
```diff
@@ -152,18 +93,22 @@
...
       <id>test.remote</id>
-      <activation>
-        <property>
-          <name>test.mode</name>
-          <value>remote</value>
-        </property>
-      </activation>
+      <properties>
+        <!-- change the build timestamp to something palatable for gcloud -->
+        <maven.build.timestamp.format>yyyyMMdd-HHmm</maven.build.timestamp.format>
```
jmcc0nn3ll: ok, note that this is different then the format used in the openjdk-runtime

meltsufin: We need to change it there as well because underscores are no longer allowed in the version tags.

tests/test-war-smoke/pom.xml
jmcc0nn3ll: Commented this execution out on latest commit, afaik this isn't available in maven but we might be able to iron out a command line to execute and drop into a maven property.

jmcc0nn3ll: > docker images | grep ${project.artifactId} | grep ${project.version} | tr -s ' ' | cut -d ' ' -f3

b124ff28bb51

I think that would grab it for the proper case...ugly though :)

meltsufin: I think I found a more reliable way to make it work:
```XML
                <configuration>
                  <executable>bash</executable>
                  <arguments>
                    <argument>-c</argument>
                    <argument>gcloud beta container images delete `docker inspect --format='{{index .RepoDigests 0}}' ${jetty.test.image}` -q</argument>
                  </arguments>
                </configuration>
```

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
       image=gcr.io/travis-app-maven-plugin/jetty_test:`date +%s`
       docker tag jetty:9.4 $image
       # do the remote test
-      gcloud docker -q -- push $image && mvn clean verify -Ptest.remote -Djetty.test.image=$image -B -q
+      gcloud docker -q -- push $image && mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean -Djetty.test.image=$image -B -q
     fi
 
 after_script:
```
meltsufin: this section is no longer needed

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
       image=gcr.io/travis-app-maven-plugin/jetty_test:`date +%s`
```
meltsufin: We no longer need this

.travis.yml
```diff
@@ -65,7 +65,7 @@ script:
...
       docker tag jetty:9.4 $image
       # do the remote test
-      gcloud docker -q -- push $image && mvn clean verify -Ptest.remote -Djetty.test.image=$image -B -q
```
meltsufin: don't pass `-Djetty.test.image=$image`

tests/test-war-smoke/pom.xml
```diff
@@ -228,35 +165,161 @@
...
+                </goals>
+                <configuration>
+                  <workingDirectory>${project.build.directory}/appengine-staging</workingDirectory>
+                  <executable>gcloud</executable>
+                  <arguments>
+                    <argument>--quiet</argument>
+                    <argument>app</argument>
+                    <argument>deploy</argument>
+                    <argument>app.yaml</argument>
+                    <argument>--promote</argument>
```
meltsufin: I don't think `--promote` is needed. In fact, it can prevent you from deleting the service.

jmcc0nn3ll: I added it because there was a previous uncleaned version that was ended up taking the test traffic, at least that is what the error messages being tossed said, and the recommendation was to use --promote.  I'll start from a clean slate and see if I can get away with not using it, the --help doc seem to imply it was on by default unless turned off for the project anyway.

meltsufin: If there is an uncleaned version, it shouldn't affect us because the URLs we construct in the tests are version-specific.

```java
      // construct the remote uri
      StringBuilder uri = new StringBuilder();
      uri.append("https://");
      uri.append(version).append("-dot-");
      if (serviceId != null) {
        uri.append(serviceId).append("-dot-");
      }
      uri.append(projectId);
      uri.append(".appspot.com/");
```

jmcc0nn3ll: It is possible the error output I was getting was because the delete was running against the running instance and it warning me I needed to push something else with --promote.  An issue was resolved by adding the app stop execution...i bet that was it

jmcc0nn3ll:
passes now however i need to tinker with a better way of cleaning local docker images that are produced by this new usage of docker.tag.long, I'll open a bug on that now because it is mostly unrelated to this PR

meltsufin:
Does this pass for you?
`mvn clean verify -Ptest.local,test.remote,test.remote.deploy,test.remote.clean`

I think there is some kind of problem with tagging the image.

jmcc0nn3ll:
Figured it out, to resolve the local test issue in travis it shifted around the order of plugin execution within the package phase.  I moved the phase of the executions in test.remote.deploy to generate-resources which should accomplish the proper ordering.  Running some local tests to make sure it works as expected and will update.

meltsufin:
Looks good! Tested on `travis_remote` branch as well.

PR #118, opened 2017-01-10
Fix Travis remote testing GCR image cleanup

meltsufin:
I just realized that the old cleanup code didn't work. It only removed the image locally, not in the GCR.


meltsufin:
Thanks! No, I don't think we need to clean the local images because each Travis build is independent.

PR #121, opened 2017-01-25
Reuse docker entrypoint from base openjdk image

### LINKED ISSUE #120: Reuse docker-entrypoint.bash from openjdk:8 image
Currently, we're overwriting the docker-entrypoint.bash file defined in the base image while duplicating most of the code in it. We should instead leverage the script in the base image. One option would be to define a different entrypoint script and call the one in the base image at the end after appending `$DEFAULT_ARGS` to `$JAVA_OPTS`.

gregw:
Fixes #120 

Move jetty configuration from an overridden docker-entrypoint.bash script
to jetty9/src/main/docker/app-env.bash.

NOT READY for merge, as this depends on https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/49 being available in the base openjdk image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,34 +76,32 @@ Once you have this configuration, you can use the Google Cloud SDK to deploy thi
      
 
 ## Entry Point Features
-The entry point for the image is [docker-entrypoint.bash](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/jetty9/src/main/docker/docker-entrypoint.bash), which does the processing of the passed command line arguments to look for an executable alternative or arguments to the default command (java).
 
-If the default command (java) is used, then the entry point sources the [setup-env.bash](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/setup-env.bash), which looks for supported features to be enabled and/or configured.  The following table indicates the environment variables that may be used to enable/disable/configure features, any default values if they are not set: 
+The [/docker-entrypoint.bash](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/docker-entrypoint.bash)
+for the image is inherited from the openjdk-runtime and it's capabilities are described in the associated 
+[README](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/README.md)
 
-|Env Var           | Description         | Type     | Default                                     |
-|------------------|---------------------|----------|---------------------------------------------|
-|`DBG_ENABLE`      | Stackdriver Debugger| boolean  | `true`                                      |
-|`TMPDIR`          | Temporary Directory | dirname  |                                             |
-|`JAVA_TMP_OPTS`   | JVM tmpdir args     | JVM args | `-Djava.io.tmpdir=${TMPDIR}`                |
-|`GAE_MEMORY_MB`   | Available memory    | size     | Set by GAE or `/proc/meminfo`-400M          |
-|`HEAP_SIZE_MB`    | Available heap      | size     | 80% of `${GAE_MEMORY_MB}`                   |
-|`JAVA_HEAP_OPTS`  | JVM heap args       | JVM args | `-Xms${HEAP_SIZE_MB}M -Xmx${HEAP_SIZE_MB}M` |
-|`JAVA_GC_OPTS`    | JVM GC args         | JVM args | `-XX:+UseG1GC` plus configuration           |
-|`JAVA_GC_LOG`     | JVM GC log file     | filename |                                             |
-|`JAVA_GC_LOG_OPTS`| JVM GC args         | JVM args | Derived from `$JAVA_GC_LOG`                 |
-|`JAVA_USER_OPTS`  | JVM other args      | JVM args |                                             |
-|`JAVA_OPTS`       | JVM args            | JVM args | See below                                   |
+This image updates the docker `CMD` and appends to the
+[setup-env.bash](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/setup-env.bash)
+script to include options and arguments to run the Jetty container, unless an executable argument is passed to the docker image.
+Addition environment variables are set including:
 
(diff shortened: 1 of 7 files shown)


jetty9/src/main/docker/Dockerfile
```diff
@@ -39,8 +39,8 @@ RUN mkdir webapps $TMPDIR \
  && chown -R jetty:jetty $JETTY_BASE $TMPDIR
 
 # Start Jetty directly
-COPY docker-entrypoint.bash /
-RUN chmod +rx /docker-entrypoint.bash
+COPY app-env.bash /
+RUN chmod +rx /app-env.bash
```
meltsufin: How about `RUN printf "\nsource /app-env.bash\n" >> /setup-env.bash`, and no modifications to the openjdk runtime?

Also, maybe rename `app-env.bash` -> `setup-env-ext.bash` in that case?

README.md
```diff
@@ -76,34 +76,32 @@ Once you have this configuration, you can use the Google Cloud SDK to deploy thi
...
-```
+
+The [/docker-entrypoint.bash](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/docker-entrypoint.bash)
+for the image is inherited from the openjdk-runtime and it's capabilities are described in the associated 
+[README](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/README.md)
+
+This image updates the docker `CMD` and appends to the
+[setup-env.bash](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/setup-env.bash)
+script to include options and arguments to run the Jetty container, unless an executable argument is passed to the docker image.
+Addition environment variables are set including:
```
meltsufin: Addition -> Additional

README.md
```diff
@@ -76,34 +76,32 @@ Once you have this configuration, you can use the Google Cloud SDK to deploy thi
...
+|`GAE_IMAGE_LABEL` |`docker.tag.long`|                                                      |
+|`JETTY_HOME`      |`jetty.home`     |`/opt/jetty-home`                                     |
+|`JETTY_BASE`      |`jetty.base`     |`/var/lib/jetty`                                      |
+|`TMPDIR`          |                 |`/tmp/jetty                                           |
+|`JETTY_ARGS`      |                 |`-Djetty.base=$JETTY_BASE -jar $JETTY_HOME/start.jar` |
+|`ROOT_WAR`        |                 |`$JETTY_BASE/webapps/root.war`                        |
+|`ROOT_DIR`        |                 |`$JETTY_BASE/webapps/root`                            |
+|`JAVA_OPTS`       |                 |`$JAVA_OPTS $JETTY_ARGS`                              |
+
+If a WAR file is found at `$ROOT_WAR`, it is unpacked to `$ROOT_DIR` if it is newer than the directory or the directory
```
meltsufin: I think this is a bit confusing. Maybe adding an example of a Dockerfile that includes a WAR and an example that includes an exploded WAR would help.

jetty9/src/main/docker/Dockerfile
```diff
@@ -39,8 +39,9 @@ RUN mkdir webapps $TMPDIR \
  && chown -R jetty:jetty $JETTY_BASE $TMPDIR
 
 # Start Jetty directly
-COPY app-env.bash /
-RUN chmod +rx /app-env.bash
+COPY setup-env-ext.bash /tmp/
+RUN cat /tmp/setup-env-ext.bash >> /setup-env.bash \
+ && rm /tmp/setup-env-ext.bash
```
meltsufin: I don't think we really need to remove the file. It's already going take up a tiny amount of space in the image because the copy is a separate layer.

jetty9/src/test/resources/structure.yaml
```diff
@@ -29,8 +29,8 @@ fileExistenceTests:
   path: '${jetty.base}/webapps'
   isDirectory: true
   shouldExist: true
-- name: 'app-env.bash exists'
-  path: '/app-env.bash'
+- name: 'setup-env.bash exists'
```
meltsufin: I think we should also add a [file contents test](https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/structure_tests/README.md#file-content-tests) to lightly verify that the append happened.

jetty9/src/main/docker/Dockerfile
```diff
@@ -39,8 +39,9 @@ RUN mkdir webapps $TMPDIR \
  && chown -R jetty:jetty $JETTY_BASE $TMPDIR
 
 # Start Jetty directly
-COPY app-env.bash /
-RUN chmod +rx /app-env.bash
+COPY setup-env-ext.bash /tmp/
+RUN cat /tmp/setup-env-ext.bash >> /setup-env.bash \
```
meltsufin: Would `#/bin/bash` at the top of `setup-env.bash` throw it off? That's the reason I kind of preferred appeding `source /setup-env-ext.bash`.

jetty9/src/main/docker/setup-env-ext.bash
```diff
@@ -1,10 +1,7 @@
 #!/bin/bash
```
meltsufin: Might be a problem if you append the whole file to `setup-env.bash`.

jetty9/src/main/docker/Dockerfile
gregw: The `#/bin/bash` is not a problem as it is still a valid comment.   However it is unnecessary and we should instead replace it with just a comment saying that the file is expected to be appended to `setup-env.bash`

jetty9/src/main/docker/Dockerfile
gregw: I still think it is worth removing it just to avoid duplication, which may confuse people looking at the image and trying to extend/debug it.   I can imagine somebody editing that file in a DockerFile and then being surprised it is not actually used.

README.md
gregw: Will add the examples... standby....

gregw: @meltsufin examples added. I think this is ready for merge now.

PR #131, opened 2017-02-15
Removing no longer needed config in the root pom.xml

### LINKED ISSUE #69: Update or delete developers in pom.xml
the pom.xml contains an incomplete developers section.
This should either be updated to be accurate or deleted and left to github to track.

meltsufin:
Fixes #69


gregw:
LGTM

PR #133, opened 2017-02-17
Remove -preview from remote test URLs and follow redirects

### LINKED ISSUE #132: Remote tests sometimes fail with HTTP Error 502 Bad gateway

meltsufin:
Fixes #132

### DIFF
--- a/tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/util/HttpUrlUtil.java
+++ b/tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/util/HttpUrlUtil.java
@@ -52,8 +52,7 @@ public final class HttpUrlUtil {
     while (waiting && System.currentTimeMillis() < expiration) {
       try {
         System.out.print(".");
-        HttpURLConnection http = (HttpURLConnection) uri.toURL().openConnection();
-        http.setRequestProperty("User-Agent", "jetty-runtime/gcloud-util-core(server-up)");
+        HttpURLConnection http = openTo(uri);
         int statusCode = http.getResponseCode();
         if (statusCode != HttpURLConnection.HTTP_OK) {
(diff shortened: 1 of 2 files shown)


tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/util/HttpUrlUtil.java
```diff
@@ -87,7 +86,6 @@ public static HttpURLConnection openTo(URI uri) throws IOException {
     log.info("HttpUrlUtil.openTo(" + uri + ")");
     HttpURLConnection http = (HttpURLConnection) uri.toURL().openConnection();
     http.setRequestProperty("User-Agent", "jetty-runtime/gcloud-util-core");
-    http.setInstanceFollowRedirects(false);
```
gregw: I still think that this test should not follow redirects, specially as we wish to test http vs https headers and there can be redirects from http to https.

Even if we do want to follow redirects, then I think it would be better to explicitly say so in the test, because it can be non obvious when trying to debug tests that redirection is happening.

meltsufin:
@gregw I had to enable redirects because it appears that when the app is just deployed, it redirects appspot.com -> preview-appspot.com. However, a few minutes later, it doesn't redirect. So, unless you turn on redirects, you cannot use either appspot.com or preview-appspot.com reliably throughout the tests.

In any caase, I will turn off redirects, like you're suggesting, but will probe both URLs.

The thing is that preview-appspot.com is supposed to go away soon. So, will want to remove the preview-appspot.com code eventually.

meltsufin:
@gregw On a second thought, I don't think it's worth the complexity of trying to connect to two different URLs. Following redirects achives the same thing essentially. The whole redirect to preview-appspot.com should go away soon anyway.

PR #138, opened 2017-03-06
Move build and test instructions to a separate document.

dlorenc:


DEVELOPING.md
```diff
@@ -0,0 +1,27 @@
...
+```console
+git clone https://github.com/GoogleCloudPlatform/jetty-runtime.git
+./scripts/cloudbuild.sh
+```
+
+## Running the Jetty image
+The resulting image is called jetty (with more specific tags also created)
+and can be run with:
+```console
+docker run gcr.io/google-appengine/jetty9jetty
```
aslo: Since this is now inside the DEVELOPING instructions, I think it might make more sense to give the local docker image name here (just `docker run jetty`), since that's what will be produced by a developer's maven build.

dlorenc: Should be fixed. Sorry, didn't mean to change that.

PR #140, opened 2017-03-07
Fix smoke services delete by adding --no-stop-previous-version to deploy

meltsufin:
Applying this workaround for builds failing due to `app services delete` failing.


meltsufin:
From recent Travis runs, it looks like the `--no-stop-previous-version` workaround didn't actually work. We'll keep it in for now though, until `services delete` is fixed.

PR #144, opened 2017-03-24
make gae module optional

gregw:
enable gae module only if GAE_INSTANCE environment variable is set.

This could perhaps be renamed to gcp, but such renaming should probably be done separately once the environment variable that identifies all GCP environments is defined.


jetty9/src/main/docker/setup-env-ext.bash
```diff
@@ -34,4 +34,9 @@ if ! type "$1" &>/dev/null; then
   export JAVA_OPTS="$JAVA_OPTS $JETTY_ARGS"
 fi
 
+# If we are deployed on a GAE platform, enable the gae module
+if [ -n "$GAE_INSTANCE" -a ! -e "$JETTY_BASE/start.d/gae.ini" ]; then
+  echo "--module=gae" > $JETTY_BASE/start.d/gae.ini
```
meltsufin: Is Jetty scanning `$JETTY_BASE/start.d/*.ini` for configuration / adding modules at startup?
Is that better than modifying `jetty.commands`? Or, could we have used `gcp.ini` instead of `jetty.commands`?

jetty9-base/src/main/jetty-config/jetty.commands
```diff
@@ -1,6 +1,6 @@
 --create-startd
 
---add-to-start=server,webapp,http,deploy,jsp,jstl,http-forwarded,resources,gae
```
meltsufin: Some of the config below like disabling the RFC7239 header should probably be part of a GCP module. Is it possible to organize it that way?

jetty9/src/main/docker/setup-env-ext.bash
gregw: Jetty's start.jar scans start.d to build it's command line at run time.
jetty.commands is used when building the image and has the ability to do --add-to-start which can download dependencies, create template config files and expand template ini files into start.d.

We could do everything by directly generating into start.d, but jetty.commands is more future proof as any future changes to configuration and/or dependencies will be picked up by the future call to --add-to-start.

In this case, I could have generated the ini file by running a start.jar --add-to-start=gae, but as it is our own module I didn't see the need for that extra level and just directly generated the ini file.

jetty9-base/src/main/jetty-config/jetty.commands
gregw: We can move those configurations to either modules/gcp.mod file as part of a fixed configuration, or we can generate them into a start.dgcp.ini file.

We loose a little niceness as by having them in jetty.commands they are used to substitute the param declarations in the appropriate ini files as they are generated.   So we will end up with the parameters in two ini files, however they will be commented out by default in the generated ini.

We can also make http-forwarded a dependency of the gcp module, so it is enabled transitively by gcp rather than explicitly here...  actually we could then use a gcp.commands file rather than directly generating the gcp.ini and the correct parameter substitution would take place.

I'll try some variations and see which is best... stand by...

meltsufin: I like the idea of a gcp module and a `gcp.commands` file that will apply to the `*.ini` Jetty config files.

meltsufin: @gregw Can you prepare the change please?

gregw:
This PR needs to be updated to used either https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/73 or https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/74 when one of them is merged.

gregw:
@meltsufin updated this PR to use the update PR previously submitted and implemented the JETTY_MODULES_ENABLED stuff we have discussed (as was kind of related to how gcp module is turned on).

README.md
```diff
@@ -15,13 +15,20 @@ The layout of this image is intended to mostly mimic the official [docker-jetty]
 Arguments passed to the `docker run` command are passed to Jetty, so the 
 configuration of the jetty server can be seen with a command like:
 ```console
-docker run launcher.gcr.io/google/jetty --list-config
+docker run gcr.io/google-appengine/jetty --list-config
```
meltsufin: I think we want to keep the `launcher` url here.

README.md
```diff
@@ -15,13 +15,20 @@ The layout of this image is intended to mostly mimic the official [docker-jetty]
...
 ```
 
 Alternate commands can also be passed to the `docker run` command, so the
 image can be explored with 
 ```console
-docker run -it --rm launcher.gcr.io/google/jetty bash
+docker run -it --rm gcr.io/google-appengine/jetty bash
```
meltsufin: Ditto on `launcher` url.

jetty9-base/src/main/jetty-base/config-scripts/jetty.commands
```diff
@@ -0,0 +1,11 @@
+--create-startd
+
+--add-to-start=server,webapp,http,deploy,jsp,jstl,resources
+
+jetty.httpConfig.outputAggregationSize=32768
```
meltsufin: Do these settings vary from Jetty defaults? Should they be part of the gcp module?

tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/test/smoke/ForwardedIntegrationTest.java
```diff
@@ -40,44 +40,6 @@
    * @throws IOException test in error
    */
   @Test
-  @LocalOnly
```
meltsufin: Why are we removing these tests?

jetty9/src/main/docker/50-jetty.bash
```diff
@@ -0,0 +1,53 @@
+# Configure the environment and CMD for the Jetty Container
```
meltsufin: There's quiet a bit of logic in this script and we need some kind of test coverage for it. I think it's possible to do it through the structure command tests. If you want to leave it to a separate PR, please file an issue for it.

README.md
gregw: OK

README.md
gregw: Ditto

jetty9-base/src/main/jetty-base/config-scripts/jetty.commands
gregw: I don't think they are really necessary anymore.  I think they are hangovers from when logging action was taken after commit.    The image should not now care and can probably use the default.

tests/test-war-smoke/src/test/java/com/google/cloud/runtimes/jetty/test/smoke/ForwardedIntegrationTest.java
gregw: Because locally now the gcp module is not enabled, nor is it's http-forwarded dependency.   We could enable it, but then we are just testing the normal jetty http-forwarded module.

jetty9/src/main/docker/50-jetty.bash
gregw: I'm keen to keep this as a single PR as there area few overlapping concerns.   But I agree that we need to test this script.   I'll add test harness in some form shortly.

gregw:
@meltsufin I added unit tests for all the functionality in the setup script... and even found a bug in the unpacking logic!

meltsufin:
@aslo Can you take a look at the structure tests @gregw added?

aslo:
Sure! Tests look good, but I think there's an issue with the use of the `--workspace` argument for the structure_tests local run. When the structure tests are run via Cloud Builder, the project root is automatically mounted at the `/workspace` directory in the container - I don't believe it's actually configurable.

So, in short, the workspace needs to refer to the project root - it would need to look like this: 
https://github.com/GoogleCloudPlatform/jetty-runtime/compare/master-gae-optional...master-gae-optional--structure-test-config?expand=1

gregw:
@meltsufin @aslo I cherry picked Alex's change into this PR.

PR #145, opened 2017-03-30
Add integration testing

aslo:
Adds runtimes_common integration tests to the jetty runtime. This is analogous to the work done for the openjdk-runtime in https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/75.


aslo:
Readme updated :)

meltsufin:
Looks good, but what about instructions on how to run the Cloud Container Runtimes Integration Tests, like [here](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/DEVELOPING.md#running-tests).
Would probably also need to clarify that these are different from the integration tests in /tests.

PR #150, opened 2017-04-11
Remove "App Engine" wording from maven module names

meltsufin:


meltsufin:
Good catch! I just removed that README altogether.

PR #152, opened 2017-04-12
Use openjdk:latest as base image

aslo:
Instead of pinning to a specific timestamp release of the openjdk runtime, let's just use openjdk:latest as a base image and force docker to pull on every build. This way we can achieve the same result without requiring a code change whenever the openjdk runtime is updated.

@meltsufin What do you think about this?


meltsufin:
@aslo  We had it this way before, but we decided that we wanted to have more reproducible builds. So, if you checkout a tag from the past and build it, you get the same resulting image. By using `:latest`, wouldn't be able to support it.
Is the only reason you proposing this is to save the additional step of updating the jetty-runtime every time we publish a new openjdk runtime?

aslo:
Ah, that's a good point. Okay, then let's drop this change.

PR #155, opened 2017-04-19
docker args

gregw:
This is a PR against #81 that removes the duplicated code from the `docker-entrypoint.bash` from [openjdk-runtime](https://github.com/GoogleCloudPlatform/openjdk-runtime).  Currently the `docker-entrypoint.bash` is replaced by this PR, but if the approach is acceptable then that file would be moved back to `openjdk-runtime` plus some workspace tests to check its functionality.

The main difference is that the `$@` shell variable is now always kept as a fully running command. The `docker-entrypoint.bash` first normalizes it (by adding or updating the java command), then each setup script can update either `$@` or `$JAVA_OPTS`.   After sourcing the setup scripts, the entry point checks if the command is still java and if it has `$JAVA_OPTS` then these are mixed in to `$@` before it is execed.


meltsufin:
Looks good, just update it after https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/86 is merged.

tests/test-war-smoke/pom.xml
```diff
@@ -45,6 +45,7 @@
           <artifactId>servlet-api</artifactId>
         </exclusion>
       </exclusions>
```
meltsufin: https://github.com/google/google-oauth-java-client/issues/145 has been fixed. So the `servlet-api` exclusion should no longer be necessary.

PR #163, opened 2017-05-09
Adds --force-delete-tags to images cleanup script.

joaoandremartins:
mvn install -Ptest.remote,test.remote.deploy,test.remote.clean was failing without it


meltsufin:
I believe they added this flag recently, and it should completely remove the image from GCR.

PR #164, opened 2017-05-09
Moves setting RUNTIME_DIR to runtime, instead of openjdk image.

joaoandremartins:
Helps fixing https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/64


joaoandremartins:
Still need to set `APP_WEB_INF_DIR` here.

meltsufin:
@joaoandremartins I think we should first submit the code change for  `format-env-appengine-vm.sh` as suggested in my last comment in GoogleCloudPlatform/openjdk-runtime#64, and then simply set `CDBG_APP_WEB_INF_DIR` here.

joaoandremartins:
This change is pending internal change 155509870 from being submitted and published.

meltsufin:
@joaoandremartins I think we need to update this PR/resolve conflicts.

joaoandremartins:
I'm waiting for the changes in the CDBG agent to be released before moving on further work here.

I noticed this morning that they haven't been published yet.

joaoandremartins:
@meltsufin @aslo PTAL

PR #168, opened 2017-05-17
Remove travis post-submit build

aslo:
There's no need for this post-submit build anymore. Let's remove some of the noise from our emails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ jdk:
 
 branches:
   only:
-  - master
   - travis_remote
 
 install:


meltsufin:
The larger question is do we need any of the travis stuff.

aslo:
I still kind of think it makes sense for PR and commit builds, as opposed to internal CI which is mostly invisible to external contributors. I propose that if the burden of maintaining the two systems becomes too much, we move off of travis entirely, but I don't think we need to do it all at once.

meltsufin:
I think we should then remove all of the `DO_REMOTE` stuff from this Travis configuration because it was only used on `master` post-submit. We can keep the Travis config for the equivalent of a local `mvn clean install`.

.travis.yml
```diff
@@ -26,40 +26,6 @@ branches:
...
-    else
-      export DO_REMOTE=false
-    fi
-  - echo DO_REMOTE = $DO_REMOTE
-  - |
-    if $DO_REMOTE; then
-      # decrypt the tavis-app-maven-plugin Cloud SDK project service account key
-      openssl aes-256-cbc -K $encrypted_ca118b112eec_key -iv $encrypted_ca118b112eec_iv -in config/travis-app-maven-plugin-caac78288652.json.enc -out config/travis-app-maven-plugin-caac78288652.json -d
-      # set Cloud SDK CLI authentication
-      gcloud auth activate-service-account --key-file config/travis-app-maven-plugin-caac78288652.json
```
meltsufin: Should delete this file too.

.travis.yml
```diff
@@ -26,40 +26,6 @@ branches:
   only:
   - travis_remote
```
meltsufin: Can remove this whole block

PR #170, opened 2017-05-17
Add support for Quickstart

### LINKED ISSUE #12: quickstart support?
Should quickstart be supported by default, or documented as a desirable feature.
Testing done in #8 using quickstart and/or annotations

cassand:
This PR allows users extending the Jetty runtime image to use Jetty quickstart and to eagerly generate quickstart-web.xml.

Currently only the existence of the quickstart script is tested but there is a possibility to create a new test application for quickstart if needed.

Fix #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,15 +57,14 @@ env_variables:
 ```
 
 ### Using Quickstart
-Jetty provide mechanisms (link to quickstart) to speed up the start time of your application by pre-scanning its content and generating configuration files.
-If you are using an extended image (link to extension) you can activate quickstart by adding this two line to your Dockerfile.
+Jetty provides mechanisms (link to quickstart) to speed up the start time of your application by pre-scanning its content and generating configuration files.
+If you are using an extended image (link to extension) you can activate quickstart by adding this line to your Dockerfile.
 
 ```dockerfile
-ENV GENERATE_QUICKSTART true
-RUN /docker-entrypoint.bash bash
+RUN /docker-entrypoint.bash /scripts/jetty/quickstart.sh
 ```
 
-This commands must appear after the command to add the war file.
+This command must appears after the command to add the war file.
 ## App Engine Flexible Environment
 When using App Engine Flexible, you can use the runtime without worrying about Docker by specifying `runtime: java` in your `app.yaml`:
 ```yaml
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Jetty provides [mechanisms](http://www.eclipse.org/jetty/documentation/9.4.x/qui
 If you are using an [extended image](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/README.md#extending-the-image) you can activate quickstart by adding this command to your Dockerfile.
 
 ```dockerfile
-RUN /docker-entrypoint.bash /scripts/jetty/quickstart.sh
+RUN /scripts/jetty/quickstart.sh
 ```
 
 This command must appears after the command to add your war file.
--- a/README.md
+++ b/README.md
@@ -57,14 +57,15 @@ env_variables:
 ```
 
 ### Using Quickstart
-Jetty provides mechanisms (link to quickstart) to speed up the start time of your application by pre-scanning its content and generating configuration files.
-If you are using an extended image (link to extension) you can activate quickstart by adding this line to your Dockerfile.
+Jetty provides [mechanisms](http://www.eclipse.org/jetty/documentation/9.4.x/quickstart-webapp.html) to speed up the start time of your application by pre-scanning its content and generating configuration files.
+If you are using an [extended image](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/README.md#extending-the-image) you can activate quickstart by adding this command to your Dockerfile.
 
 ```dockerfile
 RUN /docker-entrypoint.bash /scripts/jetty/quickstart.sh
 ```
 
-This command must appears after the command to add the war file.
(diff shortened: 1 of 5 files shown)


gregw:
If you are going to quickstart the application, then you should also bypass start.jar as that can avoid a double invocation of the JVM and also locks in place the configuration as at the time the quickstart files were generated.

So your `quickstart.sh` should also run start.jar with the --dry-run option and capture the output into a file that the entry point can source if there is a quickstart file detected.

meltsufin:
@gregw I see that `quickstart.sh` already has a line that runs `start.jar` with `--dry-run` option. 😕 

Also, can explain more about the double invocation of the JVM and bypassing `start.jar`?

Thanks!

README.md
```diff
@@ -54,6 +54,16 @@ env_variables:
   JETTY_MODULES_ENABLE: 'gzip'
 ```
 
+### Using Quickstart
+Jetty provides [mechanisms](http://www.eclipse.org/jetty/documentation/9.4.x/quickstart-webapp.html) to speed up the start time of your application by pre-scanning its content and generating configuration files.
+If you are using an [extended image](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/README.md#extending-the-image) you can activate quickstart by adding this command to your Dockerfile.
+
+```dockerfile
+RUN /docker-entrypoint.bash /scripts/jetty/quickstart.sh
```
meltsufin: @cassand I was thinking more along the lines of `RUN /scripts/jetty/quickstart.sh` and having `quickstart.sh` use `docker-entrypoint.bash` if needed. Does that not make sense?

gregw:
It does have --dry-run, but I can't see/understand what it is doing with the result

start.jar takes a little time to run and work out the module dependencies etc. Typically a few hundred ms.

If any of the modules need JVM configuration changes (eg to add a boot jar for ALPN or to set system properties in a module `[exec]` section (eg to configure JUL logging), then start.jar invokes a new instance of the JVM with the additional configuration.    Thus avoiding start.jar can avoid a second JVM startup.

We did both webapp quickstart and start.jar bypass for GAE standard and compat.

README.md
cassand: Indeed it might more convenient for the user to run directly quickstart.sh.

cassand:
Currently --dry-run is used to generate generate the classpath to then run PreconfigureQuickStartWar. 
The result of --dry-run could also be used to bypass start.jar however this would implies further change to the startup command, so I have created a separate issue ( #171 ).

@gregw  I'm also not sure to understand if bypassing start.jar is mandatory for quickstart and if not what is the impact ?

gregw:
Bypassing start.jar is not mandatory for quickstart, just a good idea as it is a quicker start and bakes in the configuration of the server that the generated quickstart files are based on.

If the image is extended and the jetty configuration changed or the app changed, then both quick start 
 and dry-run files will need to be regenerated.

README.md
```diff
@@ -56,6 +56,16 @@ env_variables:
   JETTY_MODULES_ENABLE: 'gzip'
 ```
 
+### Using Quickstart
+Jetty provides [mechanisms](http://www.eclipse.org/jetty/documentation/9.4.x/quickstart-webapp.html) to speed up the start time of your application by pre-scanning its content and generating configuration files.
```
meltsufin: Can you actually be more specific here about which files it generates?

meltsufin:
@gregw The idea is that we just want to generate the `quickstart-web.xml` during the container build time. Can you explain a bit more about the situations when this would be broken if the image is extended?

README.md
```diff
@@ -56,6 +56,16 @@ env_variables:
...
+If you are using an [extended image](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/README.md#extending-the-image) you can activate quickstart by adding this command to your Dockerfile.
+
+```dockerfile
+RUN /scripts/jetty/quickstart.sh
```
meltsufin: Maybe just add a more complete Dockerfile example?
```dockerfile
FROM launcher.gcr.io/google/jetty
ADD your-application.war $JETTY_BASE/webapps/root.war
# generate quickstart-war.xml
RUN /scripts/jetty/quickstart.sh
```

README.md
```diff
@@ -56,6 +56,16 @@ env_variables:
   JETTY_MODULES_ENABLE: 'gzip'
 ```
 
+### Using Quickstart
+Jetty provides [mechanisms](http://www.eclipse.org/jetty/documentation/9.4.x/quickstart-webapp.html) to speed up the start time of your application by pre-scanning its content and generating configuration files.
+If you are using an [extended image](https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/README.md#extending-the-image) you can activate quickstart by adding this command to your Dockerfile.
```
meltsufin: ... you can active quickstart by executing `/scripts/jetty/quickstart.sh` in your Dockerfile, after the application WAR is added.

README.md
```diff
@@ -56,6 +56,16 @@ env_variables:
...
+
+```dockerfile
+RUN /scripts/jetty/quickstart.sh
+```
+
+This command must appears after the command to add your war file.
```
meltsufin: This note wouldn't be needed anymore.

gregw:
@cassand Using the --dry-run to extract the classpath is a bit of a necessary evil.  I'm currently experimenting with a better way to achieve that for the standard Docker jetty image, so keep a watch there for improvements in the quickstart generation mechanism that will make this a bit simpler.

PR #175, opened 2017-05-22
Exports CDBG_APP_WEB_INF_DIR before 20-debug-env.bash runs.

### LINKED ISSUE #174: Cloud Debugger not configured correctly with app classpath
The problem is that `20-debug-env.bash` executes before `50-jetty.bash` that sets the `CDBG_APP_WEB_INF_DIR` env var. `CDBG_APP_WEB_INF_DIR` must be set before `20-debug-env.bash`.

I'm not sure if we can just change `50-jetty.bash` to `15-jetty.bash`.

joaoandremartins:
Fixes #174 

After a test, CDBG doesn't work yet, though --cdbg_extra_class_path is passed with the same value as before.


jetty9/src/main/docker/Dockerfile
```diff
@@ -33,12 +33,13 @@ RUN chown -R jetty:jetty $JETTY_HOME
 ENV JETTY_BASE ${jetty.base}
 ENV TMPDIR /tmp/jetty
 ADD jetty-base $JETTY_BASE
-ADD 50-jetty.bash /setup-env.d/
+ADD *.bash /setup-env.d/
```
meltsufin: I think you can do `ADD 15-cdbg-env.bash 50-jetty.bash /setup-env.d/`

jetty9/src/main/docker/15-cdbg-env.bash
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Adds bits of configuration necessary for the Cloud Debugger to work with the Jetty image.
+# It needs to run before 20-debug-env.bash from the openjdk image.
+
+# Set CDBG_APP_WEB_INF_DIR, used by CDBG in format-env-appengine-vm.sh
+export CDBG_APP_WEB_INF_DIR="${JETTY_BASE}/webapps/root/WEB-INF"
```
meltsufin: I would rename the file to `15-debug-env-jetty.bash` to protect a bit more against file collisions.

meltsufin:
Everything looks right now, but I'm still unable to hit any of my breakpoints.

joaoandremartins:
Filed internal issue 38509278, since I can't think of what else changed to cause this.
I'll take another look tomorrow with a fresh pair of eyes.

aslo:
@joaoandremartins I opened a PR into this branch that adds a test for this issue: https://github.com/GoogleCloudPlatform/jetty-runtime/pull/177

PR #178, opened 2017-05-23
Fix Cloud Debugger [WIP]

meltsufin:
I'm not sure why, but reverting the `jetty.comands` file to the way it was before logging PR was merged in, fixes the Cloud Debugger functionality.

Fixes issue #174 

@gregw Any clues?

### DIFF
--- a/jetty9-base/src/main/jetty-base/config-scripts/jetty.commands
+++ b/jetty9-base/src/main/jetty-base/config-scripts/jetty.commands
@@ -1,4 +1,3 @@
 --create-startd
---approve-all-licenses
---add-to-start=server,webapp,http,deploy,jsp,jstl,resources,logging-jul,jcl-slf4j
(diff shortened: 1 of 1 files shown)


gregw:
I expect this is because JUL logging requires a second JVM to be started and the debugger is debugging the original JVM rather than the spawned one.

meltsufin:
@gregw This sounds like a good lead. I didn't know that JUL requires a second JVM. Is that something the Jetty module does?

meltsufin:
@gregw I can't seem to find the code for the logging-jul module anywhere. Can you point me to where this whole second JVM starting is happening? Thanks!

jmcc0nn3ll:
I think it would be spawned off by the use of --exec somewhere in the jetty.command file (think that is the right file), or the vm that shows up after executing the jetty.command since I think that file was generated by a jvm call to start.jar --dry-run or something?

meltsufin:
Superseded by #183

Issue #171, opened 2017-05-18
Bypassing start.jar during quickstart


cassand:
It might be interesting to bypass the start.jar mechanisms when quickstart is enabled. However for users extending the image this would implies changing the command to start Jetty.  

There is a possibility to reuse the command to generate the classpath during quickstart (start.jar --dry-run).

Related to: #170

joakime:
Since PR #170 was merged, can this be considered closed now?

PR #183, opened 2017-05-26
Start jetty in single jvm

### LINKED ISSUE #182: Make Cloud Debugger work with logging-jul module
The issue has to do with `logging-jul` module causing Jetty to start a second JVM and the Cloud Debugger not working with this second JVM.

One option is to use `--dry-run` to precompute the Jetty start command and not rely on `start.jar` to start a child JVM.

Another option is to customize the `logging-jul` module not to use the `[exec]` option and pass `logging.properties` directly to the initial invocation of the JVM.

The first option seems better to me.

aslo:
This PR adds a dry-run of jetty at container start time. This allows us to run a single JVM process for the jetty server, even if modules with `exec` options are used. 

fixes #182


jetty9/src/main/docker/gen_start_command.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+START_COMMAND_OUT=$1
+
+java -jar -Djetty.base=${JETTY_BASE} ${JETTY_HOME}/start.jar $JETTY_ARGS --dry-run --exec-properties="${START_COMMAND_OUT}.properties" &> $START_COMMAND_OUT
```
meltsufin: Can we store the start command in an env variable rather than a file?

jetty9/src/main/docker/jetty_start.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+DIR=$(dirname $0)
+JETTY_START_COMMAND="/scripts/jetty/jetty-start-command"
+
+if [ ! -f $JETTY_START_COMMAND ]; then
+  source $DIR/gen_start_command.sh $JETTY_START_COMMAND
```
meltsufin: So, we can't generate the command with ONBUILD?

jetty9/src/main/docker/Dockerfile
```diff
@@ -35,13 +35,17 @@ ENV TMPDIR /tmp/jetty
...
  && chmod +x /setup-env.d/15-debug-env-jetty.bash \
  && chmod +x /setup-env.d/50-jetty.bash \
- && chmod +x /scripts/jetty/quickstart.sh
+ && chmod +x /scripts/jetty/quickstart.sh \
+ && chmod +x /scripts/jetty/jetty_start.sh \
+ && chmod +x /scripts/jetty/gen_start_command.sh
 
 EXPOSE 8080
-CMD ["java","-Djetty.base=${jetty.base}","-jar","${jetty.home}/start.jar"]
+CMD ["/scripts/jetty/jetty_start.sh"]
```
cassand: Some setup scripts (docker-entrypoint) in Jetty and Openjdk are expecting to see `java` in the command line to add specific arguments.  

https://github.com/GoogleCloudPlatform/jetty-runtime/blob/master/jetty9/src/main/docker/50-jetty.bash (At the end of the file )

https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/docker-entrypoint.bash (this one may not be affected)

aslo: Yep, this was something I thought about as well. I was thinking it could be an issue when executing the generated start command, since by default it includes the fully-qualified path to the java executable. This actually isn't a problem since [docker-entrypoint.sh](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk8/src/main/docker/docker-entrypoint.bash#L7) will normalize a long path to just `java`.

Also, within the `jetty_start.sh` script the generated start command is actually [passed to the docker-entrypoint](https://github.com/GoogleCloudPlatform/jetty-runtime/pull/183/files#diff-d1999af812fe2410408b449501c975b9R11), so I think this should be OK.

jetty9/src/main/docker/gen_start_command.sh
aslo: 👍  I'll look into it

jetty9/src/main/docker/jetty_start.sh
aslo: I'll also look a bit more into that option

aslo: I'm not convinced that ONBUILD is what we want here after all. An ONBUILD command would be run automatically after the FROM line in a child container's build, so any customization of jetty (i.e. via JETTY_MODULES_ENABLE) in their Dockerfile would not be reflected in the generated start command.

We could in theory generate the start command, either in the ONBUILD (or even in the jetty-runtime container), and then invalidate it if we detect that the user has made config changes that would create a new start command. But I'm not crazy about that idea, since we'd be taking on a good chunk of complexity to be able to detect those changes for marginal gains in container start time, at best. 

IMO, I think it's OK to tell users who are sensitive to container start time to run the `gen_start_command` script at the appropriate place in their dockerfiles.

jmcc0nn3ll: I think addressing container start time like this might be a premature optimization, you should gather some real world metrics related to this in this specific usage. For any application of beyond trivial size the calculation of this command line is likely to be a very minor portion of overall startup time, at least for anything I have seen thus far.  It would be unfortunate to add an option like this, document it and have users enable it thinking there would be a dramatic change and not have it materialize.

meltsufin: Agreed, we don't need to document `gen_start_command`. Just do it on startup.
@jmcc0nn3ll The reason we're doing all this is not for optimizing the start time, it's more to fix the issue with the Cloud Debugger which is caused by `start.jar` spinning up a 2nd JVM.

jmcc0nn3ll: oh absolutely, i just saw that comment about doing something for people sensitive to container start time and wanted to chime in 👍

cassand:
Also related to #171

jetty9/src/main/docker/Dockerfile
gregw: I think it would be better to leave the CMD as the jetty command and then just modify the 50-jetty.bash script to look at  the passed args and if they are for jetty, then replace them with the dry-run output.  Note however, that users should still be able to pass in args like --dry-run or --list-config and have them work correctly.    

I've proposed something similar for the standard docker image: https://github.com/appropriate/docker-jetty/pull/66

aslo: @gregw Thanks for the insight! I like that approach. I'll go ahead and make those changes to this PR.

README.md
```diff
@@ -258,6 +258,19 @@ image via a runtime mount:
 docker run -v /some-path/your-application:/app launcher.gcr.io/google/jetty  
 ```
 
+### Enabling dry-run
+The image's default start command will first run the jetty start.jar as a --dry-run to generate the JVM
+start command before starting the jetty web server. If you wish to generate the start command in your Dockerfile
+rather than at container start-time, you can run the `/scripts/jetty/gen_start_command.sh` script to generate it
```
meltsufin: `gen_start_command.sh` doesn't exist anymore

README.md
```diff
@@ -258,6 +258,19 @@ image via a runtime mount:
...
 ```
 
+### Enabling dry-run
+The image's default start command will first run the jetty start.jar as a --dry-run to generate the JVM
+start command before starting the jetty web server. If you wish to generate the start command in your Dockerfile
+rather than at container start-time, you can run the `/scripts/jetty/gen_start_command.sh` script to generate it 
+for you, i.e. 
+
+```bash
+RUN export JETTY_START_COMMAND=$(/scripts/jetty/gen_start_command.sh)
```
meltsufin: `gen_start_command.sh` doesn't exist anymore and `JETTY_START_COMMAND` is not used

jetty9/src/test/resources/structure.yaml
```diff
@@ -75,6 +79,14 @@ fileExistenceTests:
   path: '/scripts/jetty/quickstart.sh'
   isDirectory: false
   shouldExist: true
+- name: 'jetty start script exists'
+  path: '/scripts/jetty/jetty_start.sh'
+  isDirectory: false
+  shouldExist: true
+- name: 'gen_start_command.sh script exists'
+  path: '/scripts/jetty/gen_start_command.sh'
```
meltsufin: This should fail as the file doesn't exist.

jetty9/src/test/workspace/jetty-setup-start-script.bash
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+CUSTOM_START_CMD="echo 'pre-generated start command'"
```
meltsufin: This whole test is outdated.

aslo:
@meltsufin I updated to take @gregw's suggestions into consideration, and borrowed from his implementation in appropriate/docker-jetty#66. 

PTAL - this is ready for review

README.md
```diff
@@ -258,6 +258,19 @@ image via a runtime mount:
 docker run -v /some-path/your-application:/app launcher.gcr.io/google/jetty  
 ```
 
+### Enabling dry-run
+The image's default start command will first run the jetty start.jar as a --dry-run to generate the JVM
+start command before starting the jetty web server. If you wish to generate the start command in your Dockerfile
+rather than at container start-time, you can run the `/scripts/jetty/generate-jetty-start.sh` script to generate it 
+for you, i.e. 
+
+```bash
```
meltsufin: bash -> Dockerfile

aslo: Didn't know that was a thing. Nice!

aslo:
Merging now. FYI, these changes require the changes in https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/109 in order to fix #182

PR #189, opened 2017-06-09
Unify all tests into single cloudbuild execution 

aslo:
This PR refactors all of the various integration test scripts so that they can be run as a single cloudbuild execution. This has the following benefits:
* less bash!
* parallelization of build steps
* running in Cloud Container Builder guarantees repeatability and hermeticity for integration test builds
* any release candidate built with this cloudbuild configuration will necessarily have passed integration tests

Before merging, we need to do the following:
* [ ] update build scripts in internal CI to use `scripts/build.sh` with the correct args
* [x] make sure that the GCP project used for release builds will tolerate spinning up multiple app engine services on each test run


aslo:
Thanks! Documentation updates will come soon. 

Cleanup is slightly complicated since failed builds will exit immediately by default. One solution I'm considering is to have a separate cloudbuild execution that just tries to clean up known resources.

cloudbuild.yaml
```diff
@@ -3,17 +3,62 @@
...
 steps:
 # Perform maven build, omitting local docker operations
-- name: 'maven:3.3.9-jdk-8'
-  args: ['mvn', '-B', '-P-local-docker-build', '-P-test.local', '-Ddocker.tag.long=$_DOCKER_TAG', 'clean', 'install']
+- name: 'gcr.io/cloud-builders/java/mvn:3.5.0-jdk-8'
+  args:
+    - '--batch-mode'
+    - '-P-local-docker-build'
+    - '-P-test.local'
+    - '-Ddocker.tag.long=${_DOCKER_TAG}'
```
cassand: It might be judicious to override properties such as the runtime name and version in the maven configuration.

Otherwise we could publish a version that differ from the version in the tag.

aslo:
@meltsufin Updated documentation, added cleanup!

cloudbuild.yaml
aslo: Good idea!

build/Dockerfile.mvn-gcloud
```diff
@@ -0,0 +1,24 @@
+FROM maven:3.5.0-jdk-8
```
meltsufin: Maybe put a comment up top saying what this is used for?

DEVELOPING.md
```diff
@@ -24,11 +24,6 @@ TAG=my-tag
 ./scripts/build.sh gcr.io/$PROJECT_ID $TAG
```
meltsufin: Mention which permissions must be granted to the `cloudbuild.gserviceaccount.com` service account:
* App Engine Admin
* Logs Viewer
* Container Analysis Admin
* Others??
...And that the "Google Service Management API" needs to be enabled on the project.

cloudbuild.yaml
aslo: @cassand I thought about this some more, and I actually think this is OK for now. In general, the cloudbuild execution has full control over the way the image is named and tagged, so regardless of what we pass into maven, these fields are still able to be overridden. 

I agree that we should come up with a more principled way of handling these duplicated properties, but I think a better time to handle it is when we start releasing more than one image from this repository (the next major version of Java and Jetty). Is that OK with you?

cassand: Yes, as maven only download the Jetty version but does not generate any tag for the docker image it will not be disruptive for cloudbuilder to tag and push the image.

PR #195, opened 2017-06-16
Add local container tests for shutdown logging

meltsufin:
The following tasks are pending before this test will pass
- [x] Release new openjdk image
- [x] Update openjdk image in pom.xml

Related to GoogleCloudPlatform/openjdk-runtime/pull/115


meltsufin:
Yes, I did, because it wasn't working on Jenkins and the fix was not completely obvious. I created a separate issue for that #197

PR #198, opened 2017-06-23
Set APP_DESTINATION env variable in dockerfile

aslo:
Sets a variable to show where web apps should be added in child containers.


aslo:
@meltsufin Readme updated. Mind taking another look?

PR #202, opened 2017-07-28
Specify GKE zone

aslo:
Container Builder builds are failing because this gcloud configuration property was not provided.

Thanks @cassand for figuring this out!


aslo:
I added an issue to track refactoring the zone names: https://github.com/GoogleCloudPlatform/jetty-runtime/issues/203

PR #204, opened 2017-08-02
Specify gke zone, run scripts in strict mode

aslo:
This is a followup to https://github.com/GoogleCloudPlatform/jetty-runtime/pull/202


aslo:
@cassand I tried running the GKE int tests on container builder and it timed out, because the `wait.sh` script never finished. Do you know why that might be happening now that I specified the zone?

cassand:
Yes, if the application fail to deploy but the load balancer still deploy successfully the script may wait indefinitely.

PR #222, opened 2018-01-05
Use bom

### LINKED ISSUE #221: Upgrade to the latest google-cloud-logging library
We also should start using the Google Cloud Java BOM for managing library versions from Google Cloud Java.

Potential solution to #217.

ChengyuanZhao:
fixes #221

### DIFF
--- a/jetty9-base/pom.xml
+++ b/jetty9-base/pom.xml
@@ -28,6 +28,18 @@
   <artifactId>jetty9-base</artifactId>
   <packaging>jar</packaging>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bom</artifactId>
+        <version>${gcloud.api.bom.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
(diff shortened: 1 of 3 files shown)


jetty9-base/pom.xml
```diff
@@ -28,6 +28,18 @@
   <artifactId>jetty9-base</artifactId>
   <packaging>jar</packaging>
 
+  <dependencyManagement>
```
meltsufin: Why not import it in the top-level `pom.xml`?

tests/test-war-smoke/pom.xml
```diff
@@ -30,6 +30,19 @@
   <artifactId>test-war-smoke</artifactId>
   <packaging>war</packaging>
   <name>Jetty-Runtime :: Tests :: Smoke Tests (Servlet 3.1)</name>
+
+  <dependencyManagement>
```
meltsufin: You shouldn't need to import the BOM here, if it's in the top-level `pom.xml`.

jetty9-base/pom.xml
ChengyuanZhao: Will do, new to maven so I didn't know one could do that.

ChengyuanZhao:
@meltsufin changes made, PTAL

PR #231, opened 2018-04-09
Fix localhost HTTPS smoke test

meltsufin:
HTTPS testing does not apply to `localhost`. So, do not wait for the `https://localhost...` endpoint to be responsive, like we do for the remote.

### DIFF
--- a/tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/test/AbstractIntegrationTest.java
+++ b/tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/test/AbstractIntegrationTest.java
@@ -128,8 +128,10 @@ public class AbstractIntegrationTest {
   @BeforeClass
   public static void waitForServerUp() throws Exception {
     HttpUrlUtil.waitForServerUp(SERVER_URI, TIMEOUT_SECONDS, TimeUnit.SECONDS);
-    // wait for HTTPS as well
-    HttpUrlUtil.waitForServerUp(new URI(SERVER_URI.toASCIIString().replace("http:", "https:")),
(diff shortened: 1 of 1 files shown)


tests/gcloud-testing-core/src/main/java/com/google/cloud/runtime/jetty/test/AbstractIntegrationTest.java
```diff
@@ -128,8 +128,10 @@ public String getMode() {
   @BeforeClass
   public static void waitForServerUp() throws Exception {
     HttpUrlUtil.waitForServerUp(SERVER_URI, TIMEOUT_SECONDS, TimeUnit.SECONDS);
-    // wait for HTTPS as well
-    HttpUrlUtil.waitForServerUp(new URI(SERVER_URI.toASCIIString().replace("http:", "https:")),
-            TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    // wait for HTTPS as well, but only for remote
+    if (!SERVER_URI.toASCIIString().startsWith("http://localhost")) {
```
balopat: how about https://localhost? maybe `..contains("localhost")`?

meltsufin: It will always start with `http://localhost` based on the code above. If you look at the body of the `if`, we're actually replacing `http` with `https`.
I thought of just doing a `contains` but that is error-prone itself.

balopat: sounds good, thank you!

PR #250, opened 2019-04-03
Add tools

donmccasland:
Adding more build scripts.

### DIFF
--- a/scripts/deploy-check-jetty.sh
+++ b/scripts/deploy-check-jetty.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+export KOKORO_GITHUB_DIR=${KOKORO_ROOT}/src/github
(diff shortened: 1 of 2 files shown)


gregw:
@donmccasland

Can you a a more verbose description of this change? It's hard for us to support stuff we don't know what it is for.

sharifelgamal:
There isn't a whole lot to support here, these are just a couple of scripts so we can release jetty from our CI system without having to import the source code anywhere. These should never have to be touched.

gregw:
The "a couple of scripts so we can release jetty from our CI system" is all I'm after.    Just a few words to say what the changes are for... or comments in the files themselves to make them self describing would be better.   We don't want them accidentally removed in the future because there is no clear reason for them.

PR #258, opened 2019-04-30
upgrade to jetty-9.4.18

gregw:
Signed-off-by: Greg Wilkins <gregw@webtide.com>


gregw:
Ideally this would be built after https://github.com/GoogleCloudPlatform/openjdk-runtime/pull/196 has  been merged, built and released, as without that this runtime is failing on machines with 32GB of memory.
There is also an instability in the smoke tests, so I've not been able to get a clean build locally (on a new dev machine).  However, I get the same failures for the previous release, so it is not this upgrade.

gregw:
@donmccasland the smoke tests are now passing for me locally.  I cleaned up some of the test classes to give better output, but the main change was to set the heap size so the test images actually start.

PR #280, opened 2020-04-28
Review structure of builds moving scripts to a separate maven module

olamy:
Signed-off-by: olivier lamy <olamy@webtide.com>


gregw:
@olamy any idea about the CI failure?

olamy:
@gregw fixed I made more changes for docker build path context but finally change travis to force upgrade docker (per default it use an old version...)

PR #282, opened 2020-06-02
update readme with documentation on Google authentication with jetty-openid

lachlan-roberts:
Updating readme with explanation of how to use Google authentication with the jetty-openid module.


lachlan-roberts:
@googlebot I signed it!

PR #291, opened 2020-12-01
Make smoke and int tests sequential

donmccasland:
We believe smoke and int tests are conflicting with each other in some environments.  Making them sequential.


gregw (approved):
LGTM

